PR #4, opened 2024-12-06
Update README with correct public GitHub links

nemo794:
This PR updates the installation instructions in the README with the correct public GitHub links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ See `environment.yaml` for required packages.
 
 Step 1) Install the conda package manager.
 
-Step 2) Clone the `QualityAssurance` repo and `cd` into the top level directory.
+Step 2) Clone the `nisarqa` repo and `cd` into the top level directory.
 ```
-git clone git@github-fn.jpl.nasa.gov:NISAR-ADT/QualityAssurance.git
-cd QualityAssurance
+git clone git@github.com:isce-framework/nisarqa.git
+cd nisarqa
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,10 +20,10 @@ See `environment.yaml` for required packages.
 
 Step 1) Install the conda package manager.
 
-Step 2) Clone the `QualityAssurance` repo and `cd` into the top level directory.
+Step 2) Clone the `nisarqa` repo and `cd` into the top level directory.
 ```
-git clone git@github-fn.jpl.nasa.gov:NISAR-ADT/QualityAssurance.git
-cd QualityAssurance
+git clone git@github.com:isce-framework/nisarqa.git
```
gmgunter, line 25: Might be preferable to use the HTTPS URL here instead of SSH, since not all potential users will have SSH keys setup.

The repo is public so the HTTPS URL should work for everyone (though some will prefer the SSH URL).

A user that has SSH keys setup is likely to be sophisticated enough to know that they can use the SSH URL instead, whereas a user that doesn't have SSH keys setup is unlikely to know that they should use the HTTPS URL instead.

```suggestion
git clone https://github.com/isce-framework/nisarqa.git
```

gmgunter (approved):
Thanks for fixing that!

README.md
nemo794: Ah, that's a good point, and reasonable logic. Thank you!

PR #6, opened 2024-12-09
Fix precedence of code owners

gmgunter:
Later lines in a CODEOWNERS file take precedence over earlier lines. Previously, we put the two code owners on separate lines with identical wildcard patterns. This caused only the second code owner to be requested for review. This update puts both code owners on the same line, indicating co-ownership.


nemo794 (approved):
Good catch, great fix. Thank you!

PR #7, opened 2025-01-03
Update `verify_*` functions to take a `*RootParamGroup` object

nemo794:
Previously, the `verify_*` functions took a user runconfig dictionary as input. The format of this runconfig dictionary was slightly tricky for a user to construct, especially in comparison to an e.g. native Python dataclass.

This PR updates the `verify_*` functions to now take a `*RootParamGroup` dataclass as input, which is better-defined than the runconfig dictionary input, and should be more straightforward for a user to construct and pass in independently.

As a consequence, this PR also removes a chunk of copy-paste among the `verify*` functions. That copy-paste section was moved from the `verify_*` functions into `__main__.py`.


src/nisarqa/__main__.py
```diff
@@ -173,6 +173,71 @@ def load_user_runconfig(runconfig_yaml):
...
+    | nisarqa.GUNWRootParamGroup
+    | nisarqa.ROFFRootParamGroup
+    | nisarqa.GOFFRootParamGroup
+):
+    """
+    Generate a *RootParamGroup for `product_type` from a QA Runconfig yaml file.
+
+    The input runconfig file must follow the standard QA runconfig
+    format for `product_type`.
+    For an example runconfig template with default parameters (where available),
```
gmgunter, line 193: The "where available" part was a little bit hard for me to parse when we looked over the code together. I think it's actually unnecessary for understanding this and maybe it'd be best to remove it.

```suggestion
    For an example runconfig template with default parameters,
```

src/nisarqa/__main__.py
```diff
@@ -173,6 +173,71 @@ def load_user_runconfig(runconfig_yaml):
...
+    run the command line command 'nisar_qa dumpconfig <product_type>'.
+    (Ex: Use 'nisarqa dumpconfig rslc' for the RSLC runconfig template.)
+
+    Parameters
+    ----------
+    runconfig_yaml : str
```
gmgunter, line 199:
```suggestion
    runconfig_yaml : path-like
```

src/nisarqa/__main__.py
```diff
@@ -163,7 +163,7 @@ def load_user_runconfig(runconfig_yaml):
```
gmgunter, line 163: On line 162 above:

```diff
-    runconfig_yaml : str
+    runconfig_yaml : path-like
```

src/nisarqa/__main__.py
```diff
@@ -173,6 +173,71 @@ def load_user_runconfig(runconfig_yaml):
...
+        Filename (with path) to a QA runconfig yaml file for `product_type`.
+    product_type : str
+        One of: 'rslc', 'gslc', 'gcov', 'rifg', 'runw', 'gunw', 'roff', 'goff'.
+
+    Returns
+    -------
+    root_params : nisarqa.RSLCRootParamGroup or nisarqa.GSLCRootParamGroup
+                  or nisarqa.GCOVRootParamGroup or nisarqa.RIFGRootParamGroup
+                  or nisarqa.RUNWRootParamGroup or nisarqa.GUNWRootParamGroup
+                  or nisarqa.ROFFRootParamGroup or nisarqa.GOFFRootParamGroup
```
gmgunter, line 209:
```suggestion
    root_params : nisarqa.RootParamGroup
```

src/nisarqa/__main__.py
```diff
@@ -173,6 +173,71 @@ def load_user_runconfig(runconfig_yaml):
...
+        An instance of *RootParamGroup corresponding to `product_type`.
```
gmgunter, line 210:
```suggestion
        An instance of *RootParamGroup (i.e. a subtype of `nisarqa.RootParamGroup`)
        corresponding to `product_type`. For example, if `product_type` is 'gcov',
        the return type will be `nisarqa.GCOVRootParamGroup`.
```

src/nisarqa/__main__.py
```diff
@@ -173,6 +173,71 @@ def load_user_runconfig(runconfig_yaml):
     return user_rncfg
 
 
+def generate_root_params(
```
gmgunter, line 176: Currently this function (and [`build_root_params()`](https://github.com/isce-framework/nisarqa/blob/550a3c134baec89e4cba42bfcb8883a35a66c8da/src/nisarqa/parameters/rslc_caltools_params.py#L1199)) are free functions.

As an alternative, consider implementing these as class methods of the `nisarqa.RootParamGroup` class, with names such as `RootParamGroup.from_runconfig_dict()` and `RootParamGroup.from_runconfig_file()`.

(I think this should work fine without any circular dependencies, but would need to give it a try to double-check that that's the case.)

src/nisarqa/products/igram.py
```diff
@@ -15,26 +15,30 @@
...
 
     Parameters
     ----------
-    user_rncfg : nested dict
-        A dictionary whose structure matches this product's QA runconfig
-        YAML file and which contains the parameters needed to run its QA SAS.
+    root_params : nisarqa.RIFGRootParamGroup or nisarqa.RUNWRootParamGroup
+                        or nisarqa.GUNWRootParamGroup
+        Input parameters to run this QA SAS. The instance type
+        should correspond to the `product_type`.
```
gmgunter, line 38: Alternatively, we could remove the `product_type` argument and just infer the product type from the type of `root_params`. Note that the docstring should document this behavior if you decide to make this change.

Currently, `product_type` is only an argument for the `verify_igram` and `verify_offset` functions, so this would be a step towards harmonizing their interfaces with the other `verify_*` functions.

src/nisarqa/__main__.py
```diff
@@ -152,7 +152,7 @@ def dumpconfig(product_type, indent=4):
         )
 
 
-def load_user_runconfig(runconfig_yaml):
+def load_user_runconfig(runconfig_yaml: str) -> dict[str, dict]:
```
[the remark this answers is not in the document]
gmgunter: Quick follow-up about this--

I started working on a similar problem for the static layers workflow and realized that there's a flaw in our strategy here, which is that `NestedDict[str]` requires the leaf values to be strings, whereas the runconfig YAML file could contain values that are `int`s, `float`s, `list`s, etc in addition to strings.

Here's the alternative I started cooking up (disclaimer: untested):

```python
from __future__import annotations

from collections.abc import Mapping, Sequence

RunConfigScalar = str | bool | int | float | None
RunConfigList = Sequence[RunConfigScalar | "RunConfigList" | "RunConfigDict"]
RunConfigDict = Mapping[str, RunConfigScalar | RunConfigList | "RunConfigDict"]

def load_user_runconfig(runconfig_yaml: str | os.PathLike) -> RunConfigDict:
    ...
```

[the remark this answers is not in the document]
nemo794: Ah, good catch. Thank you!

For how the runconfigs are constructed for ISCE3 and QA, this should be sufficient; our keys are always `str`.

That said, I was looking to see if there were any additional edge cases. Looking at the YAML docs [here](https://yaml.org/spec/1.2.2/#10111-generic-mapping:~:text=YAML%20places%20no%20restrictions%20on%20the%20type%20of%20keys) and  [here](https://yaml.org/spec/1.2.2/#nodes:~:text=3.3.%20Representation%20Model-,3.2.1.1.%20Nodes,-A%20YAML%20node), and the [yamale docs here](https://github.com/23andMe/Yamale?tab=readme-ov-file#map---mapvalidators-keyvalidator-minint-maxint), it seems that the dictionary keys are not required to be `str` (for example, the keys could be unique `int`s).

But again, those are YAML specs, not our Runconfig specs, so either way is fine.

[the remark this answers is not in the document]
nemo794: Update:

Even with `from __future__ import annotations`, the "|" union operator was throwing errors. So, I updated them to the `Union[..]` syntax.

For reference, these were the exceptions that were raised as I worked through this issue line-by-line:

```
(qa39) niemoell@MT-316822 nisarqa % nisarqa roff_qa sam_yaml/insar.yaml
Traceback (most recent call last):
  File "/Users/niemoell/opt/anaconda3/envs/qa39/bin/nisarqa", line 5, in <module>
    from nisarqa.__main__ import main
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/__init__.py", line 150, in <module>
    from .utils._typing import *
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/utils/_typing.py", line 5, in <module>
    RunConfigScalar = str | bool | int | float | None
TypeError: unsupported operand type(s) for |: 'type' and 'type'
(qa39) niemoell@MT-316822 nisarqa % nisarqa roff_qa sam_yaml/insar.yaml
Traceback (most recent call last):
  File "/Users/niemoell/opt/anaconda3/envs/qa39/bin/nisarqa", line 5, in <module>
    from nisarqa.__main__ import main
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/__init__.py", line 150, in <module>
    from .utils._typing import *
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/utils/_typing.py", line 7, in <module>
    RunConfigList = Sequence[RunConfigScalar | "RunConfigList" | "RunConfigDict"]
TypeError: unsupported operand type(s) for |: '_UnionGenericAlias' and 'str'
(qa39) niemoell@MT-316822 nisarqa % nisarqa roff_qa sam_yaml/insar.yaml
Traceback (most recent call last):
  File "/Users/niemoell/opt/anaconda3/envs/qa39/bin/nisarqa", line 5, in <module>
    from nisarqa.__main__ import main
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/__init__.py", line 150, in <module>
    from .utils._typing import *
  File "/Users/niemoell/Desktop/qa_public/nisarqa/src/nisarqa/utils/_typing.py", line 8, in <module>
    RunConfigDict = Mapping[str, RunConfigScalar | RunConfigList | "RunConfigDict"]
TypeError: unsupported operand type(s) for |: '_UnionGenericAlias' and 'types.GenericAlias'
```

nemo794:
Thank you, @gmgunter ! Great suggestions. All notes addressed, although we should probably take a look at how [this comment about the grammar for the `RunConfigDict`](https://github.com/isce-framework/nisarqa/pull/7#discussion_r1905904810) was addressed because there were some slightly loose ends.

nemo794:
All comments addressed! Thank you, @gmgunter !

src/nisarqa/utils/typing.py
```diff
@@ -1,7 +1,15 @@
...
+    "RunConfigScalar",
+    "RunConfigList",
+    "RunConfigDict",
+]
+
 from collections.abc import Mapping, Sequence
-from typing import Union
+from typing import TypeVar, Union
+
+from nisarqa import RootParamGroup
```
gmgunter, line 12: This import is unused

```suggestion
```

src/nisarqa/utils/typing.py
```diff
@@ -1,7 +1,15 @@
 from __future__ import annotations
 
+__all__ = [
+    "RunConfigScalar",
```
gmgunter, line 4: Do we need to add `RootParamGroupT` to `__all__` as well so that we can import it as `nisarqa.RootParamGroupT`?

https://github.com/isce-framework/nisarqa/blob/0a307491b5c5d5ccbdad0601ca9c6856918bb9a4/src/nisarqa/parameters/nisar_params.py#L1538

gmgunter (approved):
LGTM, thank you!

nemo794:
Thank you, @gmgunter !

PR #8, opened 2025-01-17
Bugfix: Set `vmin` and `vmax` when plotting InSAR images

nemo794:
The new R4.0.4 sample products revealed a bug in QA. When QA re-wraps images by 7pi, the colorbar is not being generated correctly because the vmin and vmax are not being set. This impacts the PDF report and the browse images.

With the old R4.0.2 sample products, the unwrapped phase array crosses two intervals ( `[-7pi, 0)` and `[0, 7pi)`). and so it was being plotted like this:
Screenshot 2025-01-17 at 2 24 56 PM

With the new R4.0.4 sample products, the unwrapped phase array is contained within one interval ( `[0, 7pi)`). and so it was being plotted like this:
Screenshot 2025-01-17 at 2 25 58 PM

With the fix in this PR, we stretch the colorbar interval to plot correctly:
Screenshot 2025-01-17 at 2 25 21 PM

### DIFF
--- a/src/nisarqa/utils/plotting.py
+++ b/src/nisarqa/utils/plotting.py
@@ -944,8 +944,8 @@ def plot_2d_array_and_save_to_png(
     sample_spacing: Optional[tuple[float, float]] = None,
     longest_side_max: Optional[int] = None,
     *,
-    vmin: float = None,
-    vmax: float = None,
+    vmin: float | None = None,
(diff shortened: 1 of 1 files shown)


gmgunter (approved):
LGTM, thanks

nemo794:
Thank you, @gmgunter !

PR #14, opened 2025-02-11
Bugfix: Remove InSAR dtype check in product reader

nemo794:
## Background
For historical reasons, the InSAR product reader had a dtype check built into it. This check required that the QA developer know apriori the dtype of each dataset, and maintain the QA code if/when the product spec changed. This check in the product reader was intended to be a stop-gap, to be used until the actual QA XML checker was implemented and could algorithmically check the dtypes in the HDF5 file against the actual XML product specs for all datasets. That way, if/when the product spec XMLs are updated, then QA would get that update "for free".

The QA XML checker has since been implemented, and the XML vs. HDF5 dtypes are now being checked via [this function](https://github.com/isce-framework/nisarqa/blob/90dbb9e76442eda6dc34d8ed76f2225187ee22c4/src/nisarqa/utils/file_verification/checks.py#L534).

## Issue
This means that there are two independent dtype checks for the imagery datasets, and the product reader dtype for the `connectedComponents` layer has indeed fallen out of sync. In the GUNW R4.0.4 sample product, this led to a false error (aka a bug) being written in the log file.

## Solutions
There are two possible way to address:
1) Update the connected components getter to expect `uint16`, which matches the current XML product spec.
2) Remove the dtype check entirely from the reader, and let the XML Checker handle it

I lean towards option 2 because:
* Maintainability: keeps the code modular, brings the code one step towards "single responsibility", and reduces QA to having one source of truth for the dtype
* Sneakiness: The product reader does not mention dtype requirements of the datasets in the docstring; it seems strange to have it emit an error

## Discussion
I think there might be similar checks for the RSLC, GSLC, and GCOV product readers; those are potentially a bit messier (RSLC and GSLC shifted between complex32 and complex64). Would it be ok to handle them in a separate PR, or should they be updated here?

### DIFF
--- a/src/nisarqa/products/product_reader.py
+++ b/src/nisarqa/products/product_reader.py
@@ -3451,47 +3451,6 @@ class InsarProduct(NisarProduct):
         # So, treat this as a wrapper function.
         return h5_file[raster_path]
 
-    def _check_dtype(self, path: str, expected_dtype: np.dtype) -> None:
-        """
-        Check that the dataset found at `path` has the correct dtype.
-
-        Parameters
-        ----------
-        path : str
-            Path to a dataset inside the input product.
-        expected_dtype : np.dtype
-            The expected dtype for the dataset, e.g. np.complex64.
-        """
-
-        # Use lru_cache to minimize the amount of (slow) file i/o
-        @lru_cache
-        def _check_dtype_inner(path: str, expected_dtype: np.dtype) -> None:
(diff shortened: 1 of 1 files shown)


nemo794:
Tagging @xhuang-jpl and @vbrancat for awareness

gmgunter (approved):
LGTM, thank you!

nemo794:
Thank you, @gmgunter !

PR #15, opened 2025-02-11
Bugfix: Update `isJointObservation` check for InSAR products

nemo794:
Currently, QA assumes that all L1 and L2 products contain a boolean dataset `../identification/isJointObservation`.

That is true for the single-input products (RSLC, GSLC, GCOV). But for InSAR, this needs to be split into [`../identification/referenceIsJointObservation`](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/f9a912736432837ba3d05494c0b45cdefe69288a/XML/L2/nisar_L2_GUNW.xml#L35) and [`../identification/secondaryIsJointObservation`](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/f9a912736432837ba3d05494c0b45cdefe69288a/XML/L2/nisar_L2_GUNW.xml#L41).

This discrepancy is causing a false error to be written in the InSAR QA log file. This PR address that bug.

### DIFF
--- a/src/nisarqa/utils/sanity_checks.py
+++ b/src/nisarqa/utils/sanity_checks.py
@@ -213,14 +213,23 @@ def identification_sanity_checks(
         )
 
     # Verify Boolean Datasets
-    for ds_name in (
+    bool_datasets = [
         "isDithered",
         "isGeocoded",
         "isMixedMode",
         "isUrgentObservation",
-        "isJointObservation",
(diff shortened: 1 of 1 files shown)


nemo794:
Tagging @xhuang-jpl and @vbrancat for awareness

PR #16, opened 2025-02-13
Remove dtype checking for RSLC, GSLC, GCOV in product reader classes

nemo794:
In PR #14 , we removed the dtype checking for InSAR products in the product reader classes; the XML Checker rendered those checks no longer necessary, and buggy to maintain. The goal of this PR is to similarly removes the dtype checks in the RSLC, GSLC, and GCOV product readers.

Worth discussing is a proposed design change for the location of the `_get_dataset_handle()` function. Note: This function is only used during construction of the large *Raster classes for the imagery datasets. It doesn't get used by QA when accessing the tiny metadata datasets nor LUTs.

Previously, because RSLC, GSLC, and GCOV each expected different dtypes for the large imagery datasets, and because those dtypes changed over time, then `_get_dataset_handle()` was used to customize that function for each particular products, and log custom error messages accordingly.

This PR's goal was to remove those error messages. But after removing them, then most of the need for that specialization evaporated. What remained of the logic had a lot of redundancy in the child classes, with the exception of the SLCs (which need to handle the c4->c8 conversion). This meant it'd be best to refactor.

There were three approaches I considered for refactoring:
1) Maintain the current class structure, but replace the two `@abstractmethod` functions with `@staticmethod` implementations that simply did `return h5_file[raster_path]`. (Same implementation, duplicated). Then, have the `SLC()` child class override it.
2) Maintain the current class structure, but remove the two `@abstractmethod` functions, and replace with one `@staticmethod` implementation in the parent `NisarProduct` class that simply did `return h5_file[raster_path]`. Then, have the `SLC()` child class override it.
3) Remove `_get_dataset_handle()` entirely from the class structure. Instead, move it to a private helper function in the module (we have others like this), and have it always check if the dataset is `complex32`.

I went with Option 3 because it seems the most modular/reusable, but happy to chat if there's a preferred design pattern.

### DIFF
--- a/src/nisarqa/products/product_reader.py
+++ b/src/nisarqa/products/product_reader.py
@@ -293,7 +293,9 @@ def _parse_dataset_stats_from_h5(
     return raster_stats
 
 
-def _get_dataset_handle(h5_file: h5py.File, raster_path: str) -> h5py.Dataset:
+def _get_dataset_handle(
+    h5_file: h5py.File, raster_path: str
+) -> h5py.Dataset | nisarqa.ComplexFloat16Decoder:
     """
     Return a handle to the requested Dataset.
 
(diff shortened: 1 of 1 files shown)


nemo794:
Thank you, @gmgunter !

PR #21, opened 2025-04-08
Add log messages for benchmarking, etc.

nemo794:
This PR adds a decorator function that will log the start time, completion time, and arguments of a function.

This should enable easy, consistent logging for benchmarking purposes with minimal clutter and impact to existing functions.

### DIFF
--- a/src/nisarqa/products/rslc.py
+++ b/src/nisarqa/products/rslc.py
@@ -473,6 +473,8 @@ def process_backscatter_imgs_and_browse(
         "GCOV Backscatter Coefficient (gamma-0)".
         Defaults to "Backscatter Coefficient".
     """
+    log = nisarqa.get_logger()
+    log.info("Beginning processing of backscatter images...")
 
     # Select which layers will be needed for the browse image.
     # Multilooking takes a long time, but each multilooked polarization image
@@ -561,6 +563,8 @@ def process_backscatter_imgs_and_browse(
     # Construct the browse image
     product.save_browse(pol_imgs=pol_imgs_for_browse, filepath=browse_filename)
 
+    log.info("Processing complete for backscatter images.")
+
 
 # TODO - move to generic location
 def get_multilooked_backscatter_img(
@@ -595,6 +599,7 @@ def get_multilooked_backscatter_img(
         The multilooked Backscatter Image
     """
     log = nisarqa.get_logger()
+    log.info(f"Beginning multilooking for backscatter image {img.name}...")
 
     nlooks_freqa_arg = params.nlooks_freqa
     nlooks_freqb_arg = params.nlooks_freqb
@@ -651,7 +656,9 @@ def get_multilooked_backscatter_img(
             ),
         )
 
-    log.debug(f"Multilooking Image {img.name} with shape: {img.data.shape}")
+    log.debug(
+        f"Multilooking Image {img.name} with original shape: {img.data.shape}"
+    )
     log.debug(f"Y direction (azimuth) ground spacing: {img.y_axis_spacing}")
     log.debug(f"X direction (range) ground spacing: {img.x_axis_spacing}")
     log.debug(f"Beginning Multilooking with nlooks window shape: {nlooks}")
@@ -664,9 +671,8 @@ def get_multilooked_backscatter_img(
         tile_shape=params.tile_shape,
     )
 
-    log.debug(
-        f"Multilooking Complete. Multilooked image shape: {out_img.shape}"
-    )
+    log.debug(f"Final multilooked image shape: {out_img.shape}")
+    log.info(f"Multilooking complete for backscatter image {img.name}.")
 
     return out_img
 
@@ -1259,8 +1265,14 @@ def process_backscatter_and_phase_histograms(
     """
 
     # Generate and store the histograms
+    log = nisarqa.get_logger()
+    log.info("Beginning processing of backscatter and phase histograms...")
 
     for freq in product.freqs:
+        log.info(
+            "Beginning processing of backscatter and phase histograms"
+            f" for frequency {freq}..."
+        )
         generate_backscatter_image_histogram_single_freq(
             product=product,
             freq=freq,
@@ -1278,6 +1290,11 @@ def process_backscatter_and_phase_histograms(
             stats_h5=stats_h5,
             report_pdf=report_pdf,
         )
+        log.info(
+            "Processing complete for backscatter and phase histograms"
+            f" for frequency {freq}."
+        )
+    log.info("Processing complete for all backscatter and phase histograms.")
 
 
 def generate_backscatter_image_histogram_single_freq(
@@ -1328,9 +1345,7 @@ def generate_backscatter_image_histogram_single_freq(
     """
     log = nisarqa.get_logger()
 
-    log.debug(
-        f"Generating Backscatter Image Histograms for Frequency {freq}..."
-    )
+    log.info(f"Generating Backscatter Image Histograms for Frequency {freq}...")
 
     # Open one figure+axes.
     # Each band+frequency will have a distinct plot, with all of the
@@ -1360,6 +1375,7 @@ def generate_backscatter_image_histogram_single_freq(
         return power
 
     for pol_name in product.get_pols(freq=freq):
+        start_pol = time()
         with product.get_raster(freq=freq, pol=pol_name) as pol_data:
             # Get histogram probability density
             hist_density = nisarqa.compute_histogram_by_tiling(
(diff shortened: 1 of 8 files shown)


src/nisarqa/utils/utils.py
```diff
@@ -480,6 +481,60 @@ def set_logger_handler(
         log.addHandler(handler)
```
gmgunter, line 481: Consider adding a context manager function like this:

```python
from collections.abc import Generator
from contextlib import contextmanager
from datetime import datetime


@contextmanager
def log_runtime(message: str) -> Generator[None, None, None]:
    """ """
    tic = datetime.now()
    yield
    toc = datetime.now()
    nisarqa.get_logger().info(f"{message} took {toc - tic}")
```

It's a bit more flexible than the function decorator since it can be used to log the runtime of things that are not function calls. E.g. we may want to log the runtime of reading/writing individual chunks of data at some point.

Then the decorator could be implemented using the context manager, for example, like this:

```diff
-        log = nisarqa.get_logger()
-        log.info(f"Starting function: {suffix}")
-        result = func(*args, **kwargs)
-        log.info(f"Function complete: {suffix}")
-        return result
+        with log_runtime(suffix):
+            return func(*args, **kwargs)
```

src/nisarqa/utils/utils.py
```diff
@@ -480,6 +481,60 @@ def set_logger_handler(
...
+        for a in args:
+            if isinstance(a, obj_with_name_attr):
+                name = f"{a.name}. "
+                break
+        if not name:
+            for val in kwargs.values():
+                if isinstance(val, obj_with_name_attr):
+                    name = f"{val.name}. "
+                    break
+        trunc_args = tuple((trunc(i) for i in args))
```
gmgunter, line 523:
```suggestion
        trunc_args = tuple(trunc(i) for i in args)
```

src/nisarqa/utils/utils.py
```diff
@@ -480,6 +481,60 @@ def set_logger_handler(
...
+        trunc_kwargs = {key: trunc(val) for key, val in kwargs.items()}
+
+        suffix = f"`{func.__name__}`. {name}args={trunc_args}. kwargs={trunc_kwargs}."
```
gmgunter, line 526: I think the `name` thing here is a bit too kludgy.

If you're logging the runtime of a function that just has one or two salient parameters, I would use the context manager approach instead. This approach where we log all of the args & kwargs is most useful for lower-level functions with fewer parameters.

Let's remove `name` from the string here and instead have something like this:

```suggestion
        suffix = f"Called `{func.__name__}` with args={trunc_args} and kwargs={trunc_kwargs}."
```

src/nisarqa/utils/utils.py
```diff
@@ -480,6 +481,60 @@ def set_logger_handler(
...
+            max_length = 40
+            if isinstance(arg, nisarqa.YamlParamGroup):
+                max_length = 100
+            a = f"{arg!r}"
+            if len(a) < max_length or isinstance(
+                arg, (h5py.Dataset, nisarqa.ComplexFloat16Decoder)
+            ):
+                return arg
+            else:
+                return f"{a[:max_length]}[TRUNCATED]"
```
gmgunter, line 506: A more conventional way to indicate that a string has been truncated is to suffix it with `...`

```suggestion
                return f"{a[:max_length]}(...)"
```

gmgunter, line 506: Note that you could avoid the quotation marks around truncated argument strings by using string concatenation instead of f-strings (though the implementation could get a bit messy).

src/nisarqa/utils/utils.py
```diff
@@ -480,6 +481,60 @@ def set_logger_handler(
         log.addHandler(handler)
 
 
+def log_function_start_and_stop_time(func):
```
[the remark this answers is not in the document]
nemo794: Thanks, @scottstanie ! Unfortunately, `ParamSpec` was introduced in Python 3.10, and QA supports Python 3.9. Instead, will modify your suggestion to follow the `Callable[..., T]` guidance here: https://docs.python.org/3/library/typing.html#annotating-callables .

nemo794:
Comments addressed. Thank you, @gmgunter and @scottstanie ! I look forward to the next round of feedback. 🚀

src/nisarqa/products/rslc.py
```diff
@@ -1328,9 +1347,7 @@ def generate_backscatter_image_histogram_single_freq(
     """
     log = nisarqa.get_logger()
 
-    log.debug(
-        f"Generating Backscatter Image Histograms for Frequency {freq}..."
-    )
+    log.info(f"Generating Backscatter Image Histograms for Frequency {freq}...")
```
gmgunter, line 1350: The "..." seems to imply that there will be a log message later saying we've finished with the backscatter histograms, but that doesn't seem to exist.

Also, there's no similar log message for the phase histograms.

Should we remove this log message? Or else add in the missing log messages?

src/nisarqa/products/rslc.py
```diff
@@ -1623,13 +1651,16 @@ def process_range_spectra(
...
     for freq in product.freqs:
-        generate_range_spectra_single_freq(
-            product=product,
-            freq=freq,
-            params=params,
-            stats_h5=stats_h5,
-            report_pdf=report_pdf,
-        )
+        with nisarqa.log_runtime(
+            f"`generate_range_spectra_single_freq` for Freq {freq}"
```
gmgunter, line 1655:
```suggestion
            f"`generate_range_spectra_single_freq` for Frequency {freq}"
```

src/nisarqa/utils/raster_classes.py
```diff
@@ -5,6 +5,7 @@
 from collections.abc import Mapping
 from dataclasses import dataclass, fields
 from functools import cached_property
+import re
```
gmgunter, line 8: Unused import

```suggestion
```

gmgunter (approved):
LGTM, thanks!

src/nisarqa/products/rslc.py
nemo794: Added missing log messages, and set them all to logging level of "info".

src/nisarqa/utils/utils.py
```diff
@@ -480,6 +487,52 @@ def set_logger_handler(
...
+
+    See Also
+    --------
+    log_function_runtime :
+        Function decorator to log runtime of a function.
+    """
+    tic = datetime.now()
+    yield
+    toc = datetime.now()
+    nisarqa.get_logger().info(f"Runtime: {msg} took {toc - tic}")
```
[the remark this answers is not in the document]
scottstanie: Are you thinking of this just for easier parsing?

It seems like if you want more parsable log files, you might want to make the output a more naturally parse-able format [like JSONL](https://github.com/isce-framework/dolphin/blob/main/src/dolphin/_log.py#L109) . I was using this for all the unwrapping testing, where I'd log at the end of the script

```python
        elapsed = time.perf_counter() - t0
        max_mem = get_max_memory_usage(units="GB")

        # Dump extra information for the JSON logger
        extra = {
            "elapsed": elapsed,
            "max_memory_gb": max_mem,
            "unwrap_options": unwrap_option_model.model_dump(mode="json"),
        }
        logger.info(f"Completed {cur_run.name}", extra=extra)
```

then the `extra` gets dumped as a nested dict. 

the other similar thing you could do besides adding a custom level is something like https://stackoverflow.com/questions/58657285/how-does-loggings-extra-argument-work where you format the `extra` argument yourself in the `Formatter`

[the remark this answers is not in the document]
nemo794: Ah, neat ideas @scottstanie !  Unfortunately, the set of QA output files (and their extensions) has been set for a couple of years now, plus [QA inherited the format of the log file from ISCE3/SDS](https://github.com/isce-framework/nisarqa/blob/69cef7627aa031dbca6409bf212f2858f83d7831/src/nisarqa/utils/utils.py#L455-L462), and I don't really want to rock the boat this close to launch. 😬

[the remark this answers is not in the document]
gmgunter: Thanks for digging into this, @nemo794.

For posterity, I liked the idea of adding a custom logging level for runtime messages (or perhaps more broadly for messages related to performance analysis, including things like resource utilization in addition to runtimes) because:

* It would make it easy to automatically parse these messages from lots of large log files
* IMO log messages related to performance are a slightly different category than `INFO` messages (which provide basic info about the status of the process) or `DEBUG` messages (which provide detailed diagnostic info for debugging purposes), and we may find it useful now or in the future to clearly distinguish these categories
* It would make it easy to disable these runtime log messages by configuring the log level of the logger, if they start to get too verbose, without also losing the `INFO` messages

But I agree that it's best to maintain compliance with the agreed-upon log format specification for NISAR and avoid introducing a new log level. So let's ignore this suggestion for the time being.

gmgunter (approved):
LGTM, thank you!

src/nisarqa/utils/typing.py
```diff
@@ -6,6 +6,7 @@
     "RunConfigList",
     "RunConfigDict",
     "MetadataLUTT",
+    "T",
```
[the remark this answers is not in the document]
nemo794: Addressed in: https://github.com/isce-framework/nisarqa/pull/23

PR #22, opened 2025-04-14
Bugfix: Remove extraneous context manager

nemo794:
There is an extraneous context manager in the function `process_backscatter_imgs_and_browse()`. This PR removes that context manager.


gmgunter (approved):
LGTM, nice catch!

PR #25, opened 2025-04-29
Bugfix: Connected Components metrics and colorbar interval

nemo794:
Two bugs were identified when running QA on RUNW and GUNW products generated from REE "noise" data, which resulted in RUNW and GUNW products without any connected components and which resulted in layers containing all ~0.0 values.

1) Bug: Exceptions raised when computing Connected Components metrics when there are no connected components.
    - Fix: In `src/nisarqa/utils/calc.py`, default values are now set for this edge case
2) Bug: Exception raised when the pixel values imagery portion of a layer are constant (e.g. all ~0.0).
    - Fix: Log a warning, but do not raise an exception. If a layer's pixels' values are approx. constant, it should be plotted.
    - Note: There are some cases where a layer is known to be all ~zero-valued. There are existing QA checks to handle these cases.

### DIFF
--- a/src/nisarqa/utils/calc.py
+++ b/src/nisarqa/utils/calc.py
@@ -738,11 +738,12 @@ def connected_components_metrics(
     )
 
     # Percentage of pixels in largest connected component
+    percent_largest_cc = 0.0 if num_valid_cc == 0 else max(percentages_list)
     nisarqa.create_dataset_in_h5group(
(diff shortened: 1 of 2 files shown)


gmgunter (approved):
LGTM, thanks!

nemo794:
Thank you, @gmgunter !

PR #27, opened 2025-05-23
Bugfix: Unwrap longitudes for antimeridian crossings

nemo794:
### Issue
Previously, when generating the longitude and latitude values for the <LatLonQuad> in the KML files, QA was non-specific about whether the longitude values should be wrapped or unwrapped in the case of an antimeridian crossing. In practice, this resulted in:

* For L2 products, the corner points for the KML were computed [here](https://github.com/isce-framework/nisarqa/blob/30b959a7628702a05ec99ef22e0d899ef75b9266/src/nisarqa/products/product_reader.py#L1645-L1657). This resulted in lon/lat points which were unwrapped (i.e. which extended beyond +/- 180 degrees).
    * The KMLs with **unwrapped** values plotted **nicely** in Google Earth Pro.
     * For reference, the [KML documentation for LatLongQuad](https://developers.google.com/kml/documentation/kmlreference#description_15) says that the longitudes should be "in decimal degrees from 0 to ±180. (For overlays that overlap the meridian of 180° longitude, values can extend beyond that range.)".
* For L1 products, the corner points for the KML were parsed directly from the input bounding polygon ([here](https://github.com/isce-framework/nisarqa/blob/30b959a7628702a05ec99ef22e0d899ef75b9266/src/nisarqa/products/product_reader.py#L1362-L1371)). The bounding polygon's longitude values are wrapped to within +\-180 degrees; 
    * The KMLs with **wrapped** values did **not** plot nicely in Google Earth Pro. Here is a screenshot, using a KML with wrapped values (the white lines should be parallel):
Screenshot 2025-05-23 at 4 00 19 PM

### Fix

In this PR:
* Updates the `LatLonQuad` class' `__post_init__` to ensure that the corners have "unwrapped" longitude values.
* Updates code to keep lon/lat points in units of degrees throughout.
    - The KML `LatLonQuad`'s require the use of degrees. Since all of this code is for use by the KML `LatLonQuad`, it makes sense to use those units.

With this fix, we can see that the L2 GCOV KML has not changed:
Screenshot 2025-05-23 at 3 49 44 PM

And we see that the L1 RSLC  KML has been updated with unwrapped values, and now plots nicely:

Screenshot 2025-05-23 at 3 54 38 PM

### Add'l Notes

We chatted offline with the ASF DAAC to see if there was a preference for having the KML's longitude values be wrapped or unwrapped. ASF said that they post-process all values to be wrapped regardless of how they were provided. So, the decision was made to have QA's KML values be unwrapped, to better support Google Earth Pro and the KML standard.

### DIFF
--- a/src/nisarqa/utils/lonlat.py
+++ b/src/nisarqa/utils/lonlat.py
@@ -2,26 +2,61 @@ from __future__ import annotations
 
 import os
 import textwrap
+from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path
-
-import isce3
-import numpy as np
 
 import nisarqa
 
 objects_to_skip = nisarqa.get_all(__name__)
 
 
+def unwrap_longitudes(lon_lat_points: Sequence[LonLat]) -> list[LonLat]:
+    """
+    Normalize longitudes so that they form a continuous quadrilateral
+    across the antimeridian.
+
+    Arguments
+    ---------
+    lon_lat_points : Sequence of nisarqa.LonLat
+        List of nisarqa.LonLat (in degrees)
+
+    Returns
+    -------
+    unwrapped_lon_lat : list of nisarqa.LonLat
+        Copy of `lon_lat`, but in the case of an antimeridian crossing
+        the longitude values are "unwrapped" to extend beyond the
+        interval of +/-180 degrees. The ordering of the points is preserved.
+    """
+    unwrapped = [lon_lat_points[0]]
+
+    for i in range(1, len(lon_lat_points)):
+        prev_lon = lon_lat_points[i - 1].lon
+        current_lon = lon_lat_points[i].lon
+
+        delta = current_lon - prev_lon
+
+        # If it's a large jump to the west, subtract 360
+        if delta > 180:
+            current_lon -= 360
+        # If it's a large jump to the east, add 360
+        elif delta < -180:
+            current_lon += 360
+
+        unwrapped.append(LonLat(current_lon, lon_lat_points[i].lat))
+
+    return unwrapped
+
+
 @dataclass
 class LonLat:
     """
-    A point in Lon/Lat space.
+    A point in Lon/Lat space (units of degrees).
 
     Attributes
     ----------
     lon, lat : float
-        The geodetic longitude and latitude, in radians.
+        The geodetic longitude and latitude, in degrees.
     """
 
     lon: float
@@ -31,7 +66,7 @@ class LonLat:
 @dataclass
 class LatLonQuad:
     """
-    A quadrilateral defined by four Lon/Lat corner points.
+    A quadrilateral defined by four Lon/Lat corner points (in degrees).
 
     This class represents a KML gx:LatLonQuad, as described in
     https://developers.google.com/kml/documentation/kmlreference#gx:latlonquad
@@ -51,13 +86,22 @@ class LatLonQuad:
     Attributes
     ----------
     ul, ur, ll, lr : LonLat
-        The upper-left, upper-right, lower-left, and lower-right corners.
+        The upper-left, upper-right, lower-left, and lower-right corners
+        in units of degrees.
+        If there is an antimeridian crossing, longitude values will be
+        automatically unwrapped during initialization to ensure continuity.
+        For example, a quad with longitudes [179, -179] will be interpreted
+        as crossing the antimeridian and corrected to [179, 181].
     """
 
     ul: LonLat
     ur: LonLat
     ll: LonLat
     lr: LonLat
+
+    def __post_init__(self):
+        unwrapped = unwrap_longitudes((self.ul, self.ur, self.lr, self.ll))
+        self.ul, self.ur, self.lr, self.ll = unwrapped
 
 
 def write_latlonquad_to_kml(
@@ -88,15 +132,14 @@ def write_latlonquad_to_kml(
         to `output_dir`. Defaults to 'BROWSE.png'.
     """
 
-    # extract upper/lower left/right corners
-    ul, ur, ll, lr = llq.ul, llq.ur, llq.ll, llq.lr
-
-    # convert lon/lat radians to string in degrees suitable for LatLonQuad
-    ll_str = lambda c: str(np.rad2deg(c.lon)) + "," + str(np.rad2deg(c.lat))
-
+    # Construct LatLonQuad coordinates string in correct format for KML.
     # The coordinates are specified in counter-clockwise order with the first
     # point corresponding to the lower-left corner of the overlayed image.
     # (https://developers.google.com/kml/documentation/kmlreference#gx:latlonquad)
+    kml_lat_lon_quad = " ".join(
+        [f"{p.lon},{p.lat}" for p in (llq.ll, llq.lr, llq.ur, llq.ul)]
+    )
+
     kml_file = textwrap.dedent(
         f"""
         <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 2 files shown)


src/nisarqa/utils/utils.py
```diff
@@ -567,4 +573,86 @@ def load_user_runconfig(
...
+    Return successive overlapping pairs taken from the input iterable.
+
+    Example: pairwise('ABCDEFG') -> AB BC CD DE EF FG
+
+    Source: https://docs.python.org/3/library/itertools.html#itertools.pairwise
+
+    Parameters
+    ----------
+    iterable : Iterable of T
+        An Iterable of type T.
```
gmgunter, line 640:
```suggestion
    iterable : iterable of T
        The input iterable.
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+        Copy of `lons`, but unwrapped so that the absolute difference
+        between any adjacent pair of longitude values is <= 180 degrees.
+        The ordering of the points is preserved.
     """
-    unwrapped = [lon_lat_points[0]]
 
-    for i in range(1, len(lon_lat_points)):
-        prev_lon = lon_lat_points[i - 1].lon
-        current_lon = lon_lat_points[i].lon
+    # since `lons` is an Iterable, `wrap_to_interval` returns a list
```
gmgunter, line 41:
```suggestion
    # since `lons` is an iterable, `wrap_to_interval` returns an iterator
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+    lons_360 = nisarqa.wrap_to_interval(val=lons, start=-360, stop=360)
```
gmgunter, line 42:
```suggestion
    lons_360 = list(nisarqa.wrap_to_interval(val=lons, start=-360, stop=360))
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
 
-        delta = current_lon - prev_lon
+    unwrapped = [lons_360[0]]
+
+    for prev_lon, curr_lon in nisarqa.pairwise(iterable=lons_360):
```
gmgunter, line 46:
```suggestion
    for prev_lon, curr_lon in nisarqa.pairwise(lons_360):
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
-    Normalize longitudes so that they form a continuous quadrilateral
-    across the antimeridian.
+    Unwrap longitudes if the values cross the antimeridian.
+
+    Specifically, longitudes are normalized so that the absolute difference
+    between any adjacent pair of longitude values is <= 180 degrees.
+
+    Of note, in the case of the an antimeridian crossing, longitude values
+    are "unwrapped" to extend beyond the interval of +/-180 degrees
+    for the crossing.
```
gmgunter, line 26: Your description in `normalize_lon_lat_pts` below is very good. Let's use that same description here as well.

```suggestion
    Specifically, longitudes are normalized so that the absolute difference
    between any adjacent pair of longitude values is <= 180 degrees.

    All given longitude values are first wrapped to +/-360 degrees, and
    then in the case of the an antimeridian crossing they are "unwrapped"
    to extend beyond the interval of +/-180 degrees for the crossing.
```

src/nisarqa/utils/lonlat.py
gmgunter, line 42: Let's initially wrap the longitudes to [-180, 180) instead of [-360, 360).

The code will still work as long as the longitude values are within [-360, 360), but the common convention is that longitude values should be within [-180, 180] (except in the case of an antimeridian crossing).

```suggestion
    lons_180 = nisarqa.wrap_to_interval(val=lons, start=-180, stop=180)
```

gmgunter: Note this might require some updates to the docstrings as well.

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+
+    Specifically, longitudes are normalized so that the absolute difference
+    between any adjacent pair of longitude values is <= 180 degrees.
+
+    All given longitude values are first wrapped to +/-360 degrees, and
+    then in the case of the an antimeridian crossing they are "unwrapped"
+    to extend beyond the interval of +/-180 degrees for the crossing.
+
+    Arguments
+    ---------
```
gmgunter, line 74:
```suggestion
    Parameters
    ----------
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+    lon_lat_points : Iterable of nisarqa.LonLat
```
gmgunter, line 75:
```suggestion
    lon_lat_points : iterable of nisarqa.LonLat
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+        List of nisarqa.LonLat (in degrees).
+
+    Returns
+    -------
+    normalized : list of nisarqa.LonLat
+        Copy of `lon_lat`, but normalized so that the absolute difference
```
gmgunter, line 81:
```suggestion
        Copy of `lon_lat_points`, but normalized so that the absolute difference
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
+        between any adjacent pair of longitude values is <= 180 degrees.
+        The ordering of the points is preserved.
+    """
+
+    lons = unwrap_longitudes([pt.lon for pt in lon_lat_points])
+    lats = [pt.lat for pt in lon_lat_points]
```
gmgunter, line 87:
```suggestion
    lons = unwrap_longitudes(pt.lon for pt in lon_lat_points)
    lats = (pt.lat for pt in lon_lat_points)
```

src/nisarqa/utils/lonlat.py
```diff
@@ -2,52 +2,93 @@
...
-            current_lon += 360
+            curr_lon += 360
 
-        unwrapped.append(LonLat(current_lon, lon_lat_points[i].lat))
+        unwrapped.append(curr_lon)
 
     return unwrapped
 
 
+def normalize_lon_lat_pts(lon_lat_points: Iterable[LonLat]) -> list[LonLat]:
```
gmgunter, line 62: One thing I don't like about Python is that its type system makes no distinction between `Iterable`s that can be iterated over multiple times (e.g. `list`) vs `Iterable`s that can only be iterated over a single time (e.g. `Generator`).

That's a problem here because we want to iterate over `lon_lat_points` twice -- once to unwrap the longitudes and again to get the latitudes. If `lon_lat_points` was a generator, then the latitudes that we get will be an empty iterable (since we'd already exhausted the generator while unwrapping the longitudes).

In order to avoid this case, it might be prudent to specify that the input is a `Sequence` rather than an `Iterable`, since `Sequence`s can always be iterated over multiple times.

```suggestion
def normalize_lon_lat_pts(lon_lat_points: Sequence[LonLat]) -> list[LonLat]:
```

gmgunter (commented):
Few comments from our offline review

src/nisarqa/utils/utils.py
```diff
@@ -579,7 +579,7 @@ def wrap_to_interval(val: float, start: float, stop: float) -> float:
 
 def wrap_to_interval(
```
gmgunter, line 580:
```python
from typing import overload
```

```suggestion
@overload
def wrap_to_interval(
```

src/nisarqa/utils/utils.py
```diff
@@ -567,4 +573,87 @@ def load_user_runconfig(
     return user_rncfg
 
 
+def wrap_to_interval(val: float, start: float, stop: float) -> float:
```
gmgunter, line 576:
```python
from typing import overload
```

```suggestion
@overload
def wrap_to_interval(val: float, start: float, stop: float) -> float:
```

gmgunter (approved):
LGTM, thanks!

nemo794:
Woohoo! Thank you, @gmgunter !

PR #37, opened 2025-06-18
Kludge: squeeze arrays of byte strings so InSAR products do not fail

nemo794:
As of June 2025, InSAR products have a bug where `orbitFiles` is being written with shape (1, 2), instead of shape (2,). Given the prevalence of existing test data and sample products, let's have QA log and then attempt to workaround this issue to continue processing. Otherwise, all existing test datasets will cause QA to fail, which is not ideal for how close NISAR is to launch and operations.

This PR implements that kludge.

@vbrancat @xhuang-jpl


gmgunter (approved):
Thanks for the fix! LGTM

nemo794:
Thank you, @gmgunter !

PR #38, opened 2025-06-18
Copy `rfiLikelihood` for GSLC and GCOV

nemo794:
### Overview

Currently, `rfiLikelihood` is copied from the input granule to the output QA STATS.h5 file for only RSLC products. However, these `rfiLikelihood` datasets are also in the input GSLC and GCOV products. Based on offline conversations, it would be useful to users to copy this information for GSLC and GCOV as well. This PR implements that feature.

### Implementation

To implement this functionality, the following design choices were made:
* In the product reader, `get_rfi_likelihood_path()` was promoted from the child class `RSLC()` to a parent class `NonInsarProduct`. This allows GSLC and GCOV products to also have access to it. The docstring was updated accordingly.
* A new function `setup_stats_h5_non_insar_products()` was created, which first calls `setup_stats_h5_all_products()`, and then calls the additional setup functions used only by the Non-InSAR products.
    - This allows us to further refactor and reduce copy-paste in the RSLC, GSLC, and GCOV `verify_*` functions. In practice, the `save_nisar_freq_metadata_to_h5()` and `copy_non_insar_imagery_metrics()` functions can be moved from those three `verify_*` functions into one place: the new `setup_stats_h5_non_insar_products()` function.

### DIFF
--- a/src/nisarqa/utils/stats_h5_writer/setup_writer.py
+++ b/src/nisarqa/utils/stats_h5_writer/setup_writer.py
@@ -23,7 +23,7 @@ def setup_stats_h5_all_products(
     stats_h5 : h5py.File
         Handle to the output HDF5 file.
     root_params : nisarqa.typing.RootParamGroupT
-        *RootParamGroup object for the product type of `product`.
+        *RootParamGroup object corresponding to the product type of `product`.
     """
     log = nisarqa.get_logger()
     stats_file = stats_h5.filename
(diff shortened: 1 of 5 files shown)


src/nisarqa/utils/stats_h5_writer/setup_writer.py
```diff
@@ -43,6 +43,47 @@ def setup_stats_h5_all_products(
     log.info(f"Input file's runconfig copied to {stats_file}")
 
 
+def setup_stats_h5_non_insar_products(
```
gmgunter, line 46: Let's add a note to the docstring of `setup_stats_h5_all_products()` that indicates to users that this function should instead be used for non-InSAR products.

Alternatively, we could make `setup_stats_h5_all_products()` a private function and, in addition to `setup_stats_h5_non_insar_products()`, we could add a public `setup_stats_insar_products()` that is called for InSAR products. h/t @nemo794

src/nisarqa/utils/stats_h5_writer/setup_writer.py
```diff
@@ -97,6 +106,83 @@ def add_global_metadata_to_stats_h5(
...
+    """
+    Setup the STATS.h5 file for Non-InSAR products (RSLC, GSLC, GCOV).
+
+    Parameters
+    ----------
+    product : nisarqa.NonInsarProduct
+        Instance of a NonInsarProduct (RSLC, GSLC, GCOV).
+    stats_h5 : h5py.File
+        Handle to an HDF5 file where the identification metadata
+        should be saved.
```
gmgunter, line 123: What is the "identification metadata" here? It seems like this function is now doing a bunch more besides saving identification metadata.

At some point recently, we updated the docstring descriptions for `product` and `stats_h5` in `_setup_stats_h5_all_products()`:

https://github.com/isce-framework/nisarqa/blob/75673c2257f1a1d343041684370459ee1e800f33/src/nisarqa/utils/stats_h5_writer/setup_writer.py#L21-L24

Let's make a similar update here.

```suggestion
    product : nisarqa.NonInsarProduct
        The input non-InSAR (RSLC, GSLC, GCOV) product.
    stats_h5 : h5py.File
        Handle to the output HDF5 file.
```

src/nisarqa/utils/stats_h5_writer/setup_writer.py
```diff
@@ -97,6 +106,83 @@ def add_global_metadata_to_stats_h5(
...
+    """
+    Setup the STATS.h5 file for InSAR products (RIFG, RUNW, GUNW, ROFF, GOFF).
+
+    Parameters
+    ----------
+    product : nisarqa.InsarProduct
+        Instance of an InsarProduct (RIFG, RUNW, GUNW, ROFF, GOFF).
+    stats_h5 : h5py.File
+        Handle to an HDF5 file where the identification metadata
+        should be saved.
```
gmgunter, line 170:
```suggestion
    product : nisarqa.InsarProduct
        The input InSAR (RIFG, RUNW, GUNW, ROFF, GOFF) product.
    stats_h5 : h5py.File
        Handle to the output HDF5 file.
```

gmgunter (commented):
One small remaining quibble about the docstrings. Otherwise, looks good :+1:

src/nisarqa/utils/stats_h5_writer/setup_writer.py
nemo794: Good catch! I went through all of the docstrings in `nisarqa/src/nisarqa/utils/stats_h5_writer/setup_writer.py` and updated them to be consistent with this comment.

gmgunter (approved):
LGTM, thank you!

nemo794:
Thank you, @gmgunter !!

Issue #18, opened 2025-03-01
Add global metadata to the `STATS.h5` file


nemo794:
The `STATS.h5` current does not include global metadata. This should be added.

PR #39, opened 2025-06-18
Add global Attributes to STATS.h5. Refactor HDF5 writer functions.

nemo794:
This PR closes https://github.com/isce-framework/nisarqa/issues/18.

Currently, the QA STATS.h5 file does not contain global Attributes. This PR adds that metadata.

Code updates to support this feature:
* Implement new function `add_global_metadata_to_stats_h5()`, which is responsible for adding the global Attributes. Add that function to the `setup_stats_h5_all_products()` function, so that it is called during all QA SAS workflows.
* Refactor `create_dataset_in_h5group()`, so that the logic for prepping Dataset and Attribute values is modularized into a new private function, `_prep_data_for_hdf5_using_nisar_conventions()`.
* Create a new function `add_attribute_to_h5_object()`, which can be called by both `add_global_metadata_to_stats_h5()` and `create_dataset_in_h5group()` to ensure that NISAR conventions are followed for Attributes.

### DIFF
--- a/src/nisarqa/utils/utils.py
+++ b/src/nisarqa/utils/utils.py
@@ -138,13 +138,11 @@ def _prep_data_for_hdf5_using_nisar_conventions(
 ) -> ArrayLike | np.bytes_:
     """Prep data for writing to the STATS.h5 file, per NISAR conventions."""
 
-    # If a string or a list of strings, convert to fixed-length byte strings
-
-    # If `data` is an e.g. NumPy array with a numeric dtype,
-    # do not alter it.
     if isinstance(data, str) or (
         isinstance(data, Sequence) and all(isinstance(s, str) for s in data)
     ):
+        # If a string or a list of strings, convert to fixed-length byte strings
+
         # If `data` is a scalar Python (Unicode) string,
         # numpy.char.encode() returns a 0D array of byte strings.
         # If a sequence of Python strings, numpy.char.encode() returns a
@@ -173,6 +171,11 @@ def _prep_data_for_hdf5_using_nisar_conventions(
             f"`{data=}` is a sequence or array of boolean values"
             " which is not currently supported"
         )
+    else:
+        # If `data` is an e.g. NumPy array with a numeric dtype,
+        # do not alter it.
+        pass
+
     return data
 
 


src/nisarqa/utils/stats_h5_writer/setup_writer.py
```diff
@@ -43,6 +47,57 @@ def setup_stats_h5_all_products(
...
+
+    if product_type.lower() not in nisarqa.LIST_OF_NISAR_PRODUCTS:
+        raise ValueError(
+            f"{product_type=}, must one of: {nisarqa.LIST_OF_NISAR_PRODUCTS}"
+        )
+
+    product_type = product_type.upper()
+
+    global_attrs = {
+        "Conventions": "CF-1.7",
```
gmgunter, line 77: Let's remove for now since there's no known requirement for QA products to be CF-compliant, we haven't made any attempt to make them CF-compliant, and there's no science quality data contained within that end users will need to use with CF-compatible software.

```suggestion
```

src/nisarqa/utils/utils.py
```diff
@@ -207,58 +257,68 @@ def create_dataset_in_h5group(
...
+        Name (key) for the new Attribute.
+    attr_value : array_like or str or bool
+        The data to be stored as the new Attribute's value.
+
+    See Also
+    --------
+    create_dataset_in_h5group :
+        To create a new Dataset (including Attributes) in the HDF5 file.
+        Please use that function to ensure all NISAR conventions are followed
+        regarding add a `description`, `units`, etc.
```
gmgunter, line 311:
```suggestion
        Please use that function to ensure all NISAR conventions are followed,
        including adding a `description`, `units`, etc.
```

src/nisarqa/utils/utils.py
```diff
@@ -207,58 +257,68 @@ def create_dataset_in_h5group(
...
+
+    Notes
+    -----
+    Please supply Python strings for arguments. This function handles the
+    conversion to fixed-length byte strings to meet ISCE3 conventions for R4.
```
gmgunter, line 316:
```suggestion
    conversion to fixed-length byte strings to meet ISCE3 conventions for R4.0+.
```

src/nisarqa/utils/utils.py
```diff
@@ -133,6 +133,50 @@ def raise_(exc):
...
 
 
+def _prep_data_for_hdf5_using_nisar_conventions(
+    data: ArrayLike | str | bool,
+) -> ArrayLike | np.bytes_:
+    """Prep data for writing to the STATS.h5 file, per NISAR conventions."""
+
+    # If a string or a list of strings, convert to fixed-length byte strings
+
+    # If `data` is an e.g. numpy array with a numeric dtype,
```
gmgunter, line 143:
```suggestion
    # If `data` is an e.g. NumPy array with a numeric dtype,
```

src/nisarqa/utils/utils.py
```diff
@@ -133,6 +133,50 @@ def raise_(exc):
...
+        or np.issubdtype(data.dtype, np.str_)
+    ):
+        raise NotImplementedError(
+            f"`{data=}` and has dtype `{data.dtype}`, which is not"
+            " currently supported. Suggestion: Make `ds_data` a list or tuple"
+            " of Python strings, or an ndarray of fixed-length byte strings."
+        )
+    elif isinstance(data, bool):
+        data = "True" if data else "False"
+        data = np.bytes_(data)
```
gmgunter, line 167:
```suggestion
        data = np.bytes_("True" if data else "False")
```

src/nisarqa/utils/utils.py
```diff
@@ -133,6 +133,50 @@ def raise_(exc):
...
+) -> ArrayLike | np.bytes_:
+    """Prep data for writing to the STATS.h5 file, per NISAR conventions."""
+
+    # If a string or a list of strings, convert to fixed-length byte strings
+
+    # If `data` is an e.g. numpy array with a numeric dtype,
+    # do not alter it.
+    if isinstance(data, str) or (
+        isinstance(data, Sequence) and all(isinstance(s, str) for s in data)
+    ):
```
gmgunter, line 147: The order of these comments looks weird.

I think the first comment should be attached to the `if` statement here, and the second comment should perhaps go to the end of the function, before we `return data`.

nemo794:
Thank you, @gmgunter ! All comments addressed. I look forward to your re-review!

PR #42, opened 2025-06-24
Move RFI data into `sourceData`

nemo794:
Previously, the RFI information was being copied from the input RSLC/GSLC/GCOV granule into the STATS.h5 Group `/science/LSAR/RFI`, like this:
Screenshot 2025-06-05 at 4 22 36 PM

However, since this is a verbatim copy of information from the input granule (aka Source Data), and because it is not a part of CalTools, let's instead copy it into the `sourceData` Group, like this:

Screenshot 2025-06-05 at 4 49 21 PM

Expanded:
Screenshot 2025-06-05 at 4 49 12 PM


gmgunter:
We discussed this offline and ultimately decided against moving the RFI group:

* Users generally prefer a flatter subdirectory structure to make things easier to find
* Along the same lines, users of this product probably don't care whether the QA metrics it contains were copied from the source data or computed directly by the QA code -- we could consider that an implementation detail
* There are some users of the QA products currently for whom this update might be a breaking change

PR #43, opened 2025-06-26
Bugfix for NumPy 2.0: Integer overflow for Connected Components labels

nemo794:
When creating the `boundaries` values to generate the discrete colorbar for the Connected Components plot, an integer overflow is occurring when using NumPy 2.0. This PR handles that bug.

h/t @gmgunter for identifying the underlying cause:
```
Yeah this changed in NumPy 2.0. It wasn't a bug in older NumPy versions -- the datatype would automatically be promoted to avoid overflow.
[1] https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion
[2] https://numpy.org/neps/nep-0050-scalar-promotion.html
```

### DIFF
--- a/src/nisarqa/utils/plotting.py
+++ b/src/nisarqa/utils/plotting.py
@@ -3561,14 +3561,14 @@ def plot_connected_components_layer(
             " permissible interval for QA's plotting tools. Please update QA."
         )
     else:
-        labels_modified_dtype = labels.astype(np.int64)
+        labels_promoted = labels.astype(np.int64)
 
     # This assumes that `labels` is in sorted, ascending order.
     boundaries = np.concatenate(
         (
-            [labels_modified_dtype[0] - 1],
-            labels_modified_dtype[:-1] + np.diff(labels_modified_dtype) / 2.0,
-            [labels_modified_dtype[-1] + 1],
+            [labels_promoted[0] - 1],
+            labels_promoted[:-1] + np.diff(labels_promoted) / 2.0,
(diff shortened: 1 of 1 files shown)


src/nisarqa/utils/plotting.py
```diff
@@ -3534,15 +3534,44 @@ def plot_connected_components_layer(
...
+    # `boundaries` variable, we'll run into overflow errors.
+    # Prior to NumPy 2.0 (e.g. NumPy 1.26), the datatype would automatically
+    # be promoted to avoid overflow. Starting with NumPy 2.0, datatypes were
+    # no longer promoted. So, manually promote the datatype to avoid overflow.
+    # Sources:
+    #     [1] https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion
+    #     [2] https://numpy.org/neps/nep-0050-scalar-promotion.html
+    if (np.nanmin(labels) <= (-(2**63)) + 1) or (
+        np.nanmax(labels) >= (2**63 - 1) - 1
+    ):
```
gmgunter, line 3553: Let's use [`np.iinfo()`](https://numpy.org/doc/2.2/reference/generated/numpy.iinfo.html) here

```suggestion
    info = np.iinfo(np.int64)
    if (np.min(labels) <= info.min) or (np.max(labels) >= info.max):
```

nemo794:
Thank you, @gmgunter !

PR #44, opened 2025-07-16
Bugfix: `fmod` to modulo operator for wrapping

nemo794:
As noted by @gmgunter , there is a bug when wrapping a value into an interval:

```python
>>> import nisarqa
>>> nisarqa.wrap_to_interval(-200.0, start=-180.0, stop=180.0)
-200.0
```

The [`math.fmod`](https://docs.python.org/3/library/math.html#math.fmod) function computes the remainder with the same side as the left-hand side operand, whereas Python's `%` (modulo) operator computes the remainder with the same sign as the right-hand side operand.

For our `wrap_to_interval()`, we want the result to be positive (see comment in the code), so we need to update the code to use Python's modulo operator.

This PR implements that change.

### DIFF
--- a/src/nisarqa/utils/utils.py
+++ b/src/nisarqa/utils/utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import logging
-import math
 import os
 import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


gmgunter (approved):
Thanks for the fix!

PR #46, opened 2025-07-24
Bugfix: XML Checker - Account for UTM vs. polar stereo projections

nemo794:
### Background

As noted by @vbrancat in on the internal repo (https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/issues/242), georeferencing "projection" Datasets are required by CF-1.7 Conventions to contain specific Attributes. Different projections have different required Attributes, and there are some Attributes in common.

NISAR products use either UTM or polar stereographic projections, and each "projection" Dataset should contain the appropriate projection-specific parameters. For simplicity, ADT decided that the XMLs should only contain the Attributes for UTM, and then the corresponding DOCX/PDF product specification documentation would explain the polar-stereo Attributes separately.

### Current QA Status

Currently, the XML product specification documents and the QA XML Checker always assume granules are in UTM. This means that if a polar-stereo L2 product is QA'ed, then QA will incorrectly log several ERRORs.

### What this PR Accomplishes

The goals of this PR are:
* Quiet those false errors
* Robustly check for the presence of the correct projection-specific Attributes
* Checks that "projection" Datasets in the XML contain the necessary Attributes for UTM and NISAR conventions

### DIFF
--- a/src/nisarqa/validate/xml_spec_checker/policy.py
+++ b/src/nisarqa/validate/xml_spec_checker/policy.py
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
     return {"lang", "app"}
 
 
+def common_nisar_projection_attributes() -> set[str]:
+    """
+    Attributes common to NISAR "projection" Datasets but not required by CF-1.7.
+
+    These Attributes have been added to all "projection" Datasets by
+    NISAR convention. However, they are not required by CF-1.7 conventions
+    for either UTM or polar-stereographic projections.
+
+    Returns
+    -------
+    set[str]
+        The set of Attributes that are common to NISAR "projection" Datasets
+        but not required by CF-1.7 for UTM nor polar-steregraphic projections.
+    """
+    return {
+        "ellipsoid",
+        "epsg_code",
+        "longitude_of_projection_origin",
+        "spatial_ref",  # in the future, `spatial_ref` might update to `crs_wkt`
+    }
+
+
+def common_cf17_projection_attributes() -> set[str]:
+    """
+    CF-1.7 required Attributes common to UTM and polar-stereo "projection" Datasets.
+
+    These are per CF-1.7 Conventions:
+    https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html
+
+    Returns
+    -------
+    set[str]
+        The set of CF-1.7 required Attributes common to all "projection" Datasets.
+    """
+    return {
+        "grid_mapping_name",
+        "false_easting",
+        "false_northing",
+        "semi_major_axis",
+        "inverse_flattening",
+        "latitude_of_projection_origin",
+    }
+
+
+def unique_cf17_utm_attributes() -> set[str]:
+    """
+    CF-1.7 required Attributes unique to UTM "projection" Datasets.
+
+    These are per CF-1.7 Conventions:
+    https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html
+
+    Returns
+    -------
+    set[str]
+        The set of required Attributes unique to UTM "projection" Datasets.
+    """
+    return {
+        "utm_zone_number",
+        "scale_factor_at_central_meridian",
+        "longitude_of_central_meridian",
+    }
+
+
+def unique_cf17_polar_stereo_attributes() -> set[str]:
+    """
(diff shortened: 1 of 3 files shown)


src/nisarqa/validate/xml_spec_checker/policy.py
```diff
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
...
+    """
+    Attributes common to NISAR "projection" Datasets but not required by CF-1.7.
+
+    These Attributes have been added to all "projection" Datasets by
+    NISAR convention. However, they are not required by CF-1.7 conventions
+    for either UTM or polar-stereographic projections.
+
+    Returns
+    -------
+    set[str]
```
gmgunter, line 220: Typically we would document types like this more informally in the docstring (since NumPy docstring conventions predate this style of type annotation)

```suggestion
    set of str
```

src/nisarqa/validate/xml_spec_checker/policy.py
```diff
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
...
+def common_cf17_projection_attributes() -> set[str]:
+    """
+    CF-1.7 required Attributes common to UTM and polar-stereo "projection" Datasets.
+
+    These are per CF-1.7 Conventions:
+    https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html
+
+    Returns
+    -------
+    set[str]
```
gmgunter, line 241:
```suggestion
    set of str
```

src/nisarqa/validate/xml_spec_checker/policy.py
```diff
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
...
+def unique_cf17_utm_attributes() -> set[str]:
+    """
+    CF-1.7 required Attributes unique to UTM "projection" Datasets.
+
+    These are per CF-1.7 Conventions:
+    https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html
+
+    Returns
+    -------
+    set[str]
```
gmgunter, line 263:
```suggestion
    set of str
```

src/nisarqa/validate/xml_spec_checker/policy.py
```diff
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
...
+def unique_cf17_polar_stereo_attributes() -> set[str]:
+    """
+    CF-1.7 required Attributes unique to polar stereo "projection" Datasets.
+
+    These are per CF-1.7 Conventions:
+    https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/apf.html
+
+    Returns
+    -------
+    set[str]
```
gmgunter, line 282:
```suggestion
    set of str
```

src/nisarqa/validate/xml_spec_checker/policy.py
```diff
@@ -207,6 +207,91 @@ def ignored_xml_annotation_attributes() -> set[str]:
...
+        The set of required Attributes unique to polar stereographic
+        "projection" Datasets.
+    """
+    return {
+        "straight_vertical_longitude_from_pole",
+        # CF-1.7 requires either "scale_factor_at_projection_origin" or
+        # "standard_parallel".
+        # ISCE3's outputs use "standard_parallel", so for now simply use that.
+        "standard_parallel",
```
gmgunter, line 291: According to [this](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-grid-mappings:~:text=Either%20standard_parallel%20(EPSG%209829)%20or%20scale_factor_at_projection_origin%20(EPSG%209810)), `standard_parallel` is required for EPSG 9829 and `scale_factor_at_projection_origin` is required for EPSG 9810.

ISCE3 uses EPSG 3413 for Polar Stereographic North and EPSG 3031 for Polar Stereographic South.

If you look at the WKT2 string for e.g. [EPSG 3413](https://epsg.io/3413.wkt2), it references EPSG 9829 (`"Polar Stereographic (variant B)"`), so I think we're doing the right thing by including `standard_parallel` in this case.

gmgunter (approved):
LGTM, thanks!

PR #48, opened 2025-08-01
Repo Reorg: Modularize InSAR plotting functions

nemo794:
### Overview

This PR reorganizes the InSAR plotting functions from all existing in one large Python module (`src/nisarqa/utils/plotting.py`, which was 4,153 lines) into a series of smaller Python modules located under the new directory `src/nisarqa/plotting/`.  The breakdown is based primarily on the name of the raster layer being plotted, although in the case of side-by-side plots then the functions for both raster layers would be in a single module.

### Changes in this PR

There are no changes to the code functionality itself:
* All classes and functions were copy-paste; no changes
* The `__init__.py`s and import statements were updated to reflect the new structure

### Recommended Follow-on PR

After this PR is reviewed and merged, it would be advised to refactor `src/nisarqa/products/rslc.py` to take advantage of this new `processing` group. When that happens, we might decide to add subdirectories of `../processing/insar` and `../processing/non_insar`.


src/nisarqa/processing/__init__.py
```diff
@@ -0,0 +1,14 @@
+from .utils_for_plotting import *
+from .utils_for_processing import *
```
gmgunter, line 2: Consider these alternative names for brevity

```suggestion
from .plotting_utils import *
from .processing_utils import *
```

gmgunter (approved):
Looks great, thank you!

PR #49, opened 2025-08-11
Repo Reorg: Modularize `caltools.py`

nemo794:
### Overview

This PR reorganizes the Calibration Tools functions from one large Python module (`src/nisarqa/products/caltools.py`, which was 1,993 lines) into a series of smaller Python modules located under the new subdirectory `src/nisarqa/processing/caltools/`.  There is one new module per CalTool, plus a separate module for PTA plotting functionality.

### Changes in this PR

There are no changes to the code functionality itself:
* All classes and functions were copy-paste; no changes
* The `__init__.py`s and import statements were updated to reflect the new structure

These updates were tested using a RSLC HDF5 that contained corner reflectors. The output QA STATS.h5 and REPORT.pdf were confirmed to contain valid outputs from the tools.

These updates were not tested with GSLC HDF5, due to lack of readily-accessible test dataset. Happy to pursue this, if requested. Thanks!


src/nisarqa/processing/caltools/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .abscal import *
+from .noise_equivalent_backscatter import *
+from .pta import *
+from .pta_plotting import *
```
gmgunter, line 4: FYI there's an inconsistency here w.r.t a previous PR -- unlike `src/nisarqa/input_product_readers/__init__.py`, there's no comment here noting the lack of `from ._utils import *`.

https://github.com/isce-framework/nisarqa/blob/40c1112a37fd1264d672fe8702392fada4dcd0a3/src/nisarqa/input_product_readers/__init__.py#L15-L19

IMO we should just remove that comment. It's not providing much value, and it seems tedious to copy-paste it every time we introduce an `_utils.py` module in a new subdirectory.

gmgunter (approved):
LGTM, thank you!

PR #56, opened 2025-08-18
Bugfix: Copy `rfilikelihood` correctly for RSLC

nemo794:
Follow-on to #38, where there must have been a incorrectly-resolved merge conflict.

### Issue

In brief, QA is trying to copy the `rfiLikelihood` Datasets from the input RSLC HDF5 to the QA STATS HDF5 **twice**.

It is using `copy_rfi_metadata_to_stats_h5()` to copy `rfiLikelihood` inside these functions:
* `rslc_qa()`
* `setup_stats_h5_non_insar_products()` (which is called directly inside `rslc_qa()`)

This was not detected previously due to a `try..except` block treating all `RuntimeError` the same; this allowed the QA bug to be quietly logged in the log file.

### Fix

In this PR, we:
* Remove the redundant call to `copy_rfi_metadata_to_stats_h5()`
* Make the `try..except` specific to the error messages

### DIFF
--- a/src/nisarqa/processing/stats_h5_writer/setup_writer.py
+++ b/src/nisarqa/processing/stats_h5_writer/setup_writer.py
@@ -277,16 +277,26 @@ def copy_rfi_metadata_to_stats_h5(
                 )
                 try:
                     in_file.copy(src_path, stats_h5, dest_path)
-                except RuntimeError:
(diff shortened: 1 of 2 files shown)


nemo794:
Thank you, @gmgunter !

PR #66, opened 2025-08-26
Bugfix: cleanup for gcov rgb images

nemo794:
This PR addresses a bug that was introduced in #50 , which is when we removed product-specific namespaces. The bug only appears for GCOV granules where multiple terms are combined to form an RGB browse image. (It does not appear for e.g. the NISAR sample products, which only contain single pol Freq A HHHH, which generate a greyscale browse image.)

These bugfixes were tested on a GCOV dataset generated using multiple polarizations in frequency A; this is the resulting RGB browse image:

BROWSE

### DIFF
--- a/src/nisarqa/input_product_readers/gcov_reader.py
+++ b/src/nisarqa/input_product_readers/gcov_reader.py
@@ -316,13 +316,13 @@ class GCOV(NonInsarGeoProduct):
             )
 
             for gray_img in pol_imgs.values():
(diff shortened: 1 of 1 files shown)


gshiroma (approved):
Thank you, @nemo794 . It works with the dataset I was testing.  I appreciate the quick fix.

nemo794:
Thank you for finding the bug, and for the quick testing, @gshiroma !

nemo794:
Thank you, @gmgunter !

PR #67, opened 2025-08-29
Bugfix: GOFF Quiver Plots -- Arrows plot on X/Y grid

nemo794:


src/nisarqa/raster_classes.py
```diff
@@ -462,19 +462,25 @@ class GeoRaster(SARRaster):
...
     """
 
     # Attributes of the input array
     x_spacing: float
-    x_start: float
-    x_stop: float
-
     y_spacing: float
-    y_start: float
-    y_stop: float
```
gmgunter, line 474: `x_start` and friends are still attributes of the dataclass, but we're not going to pass them to the `__init__` method anymore. Let's use `dataclasses.field` to tell Python that these aren't going to be passed as arguments to the `__init__` method that it synthesizes for you, as demonstrated below:

```python
from dataclasses import field

@dataclass
class GeoRaster(SARRaster):

    x_start: float = field(init=False)
    x_stop: float = field(init=False)

    y_start: float = field(init=False)
    y_stop: float = field(init=False)
```

src/nisarqa/raster_classes.py
```diff
@@ -499,6 +505,30 @@ def x_axis_limits(self) -> tuple[float, float]:
...
+    def x_start(self) -> float:
+        return float(self.x_coordinates[0])
+
+    @property
+    def x_stop(self) -> float:
+        # X in meters (units are specified as meters in the product spec)
+        # For NISAR, geocoded grids are referenced by the upper-left corner
+        # of the pixel to match GDAL conventions. So add the distance of
+        # the pixel's side to far right side to get the actual stop value.
+        return float(self.x_coordinates[-1] + self.x_axis_spacing)
```
gmgunter, line 518: Oops, we were wrong about this. Let's fix this in a future PR.

The subtlety here is that `isce3.product.GeoGridParameters` is referenced to the upper-left corner of the first pixel, *but* the `xCoordinates` and `yCoordinates` datasets that we store in NISAR products don't follow this convention. Instead, they represent the center of each pixel in the grid.

(See here for reference: https://github.com/gmgunter/isce3/blob/e4e4e568487d4e6c37f63c323eace51e723e4397/python/packages/nisar/static/product.py#L382-L395)

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
+    -------
+    out : numpy.ndarray
+        Copy of `arr` that has been decimated along the first two
+        dimensions to have approx. square pixels.
+    ky, kx : int
+        The stride used for performing decimation in the X and Y directions,
+        respectively.
     """
-    arr = raster_obj.data[...]
+    arr = np.copy(arr)
```
gmgunter, line 766: Don't think we need to make a copy of the input array since we're not modifying the input array (hopefully...)

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
+        approx. square pixels.
+    y_axis_spacing : float
+        Pixel Spacing in Y direction (azimuth for range-Doppler rasters).
+    x_axis_spacing : float
+        Pixel Spacing in X direction (range for range-Doppler rasters).
+
+    Returns
+    -------
+    out : numpy.ndarray
+        Copy of `arr` that has been decimated along the first two
```
gmgunter, line 760:
```suggestion
        View of `arr` that has been decimated along the first two
```

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
+        dimensions to have approx. square pixels.
+    ky, kx : int
+        The stride used for performing decimation in the X and Y directions,
```
gmgunter, line 763:
```suggestion
        The stride used for performing decimation in the Y and X directions,
```

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
+        approx. square pixels.
+    y_axis_spacing : float
+        Pixel Spacing in Y direction (azimuth for range-Doppler rasters).
+    x_axis_spacing : float
+        Pixel Spacing in X direction (range for range-Doppler rasters).
+
+    Returns
+    -------
+    out : numpy.ndarray
+        Copy of `arr` that has been decimated along the first two
```
gmgunter, line 731:
```suggestion
        View of `arr` that has been decimated along the first two
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +566,282 @@ def truncate_colormap(
...
+    for i in range(num_arrows_y_direction):
+        for j in range(num_arrows_x_direction):
+            aztime = target_in_rdr_matrix[i, j, 0]
+            srange = target_in_rdr_matrix[i, j, 1]
+            # Skip NaN pixels (this is usually geocoding fill)
+            if not np.isfinite(aztime) or not np.isfinite(srange):
+                x_tail_coords_shifted[i, j] = np.nan
+                y_tail_coords_shifted[i, j] = np.nan
+
+            else:
```
seongsujeong, line 803:
```suggestion
            continue

            xyz = isce3.geometry.rdr2geo_bracket(
```
Early return

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
         Copy of raster_obj.data array that has been decimated along the
-        first two dimensions to have square pixels.
+        first two dimensions to have approx. square pixels.
+    ky, kx : int
+        The stride used for performing decimation in the X and Y directions,
+        respectively.
+    """
+    # Decimate to square pixels.
+    return decimate_array_to_square_pixels_with_strides(
+        arr=raster_obj.data[...],
```
gmgunter, line 706: `raster_obj.data` has type `numpy.typing.ArrayLike`, which doesn't guarantee that we can perform this operation (`raster_obj.data[...]`). All we know is that we can convert it to a NumPy array.

Let's ensure that this always returns a copy, because functions sometimes return a copy and sometimes return a view can be tricky to use safely.

```suggestion
        arr=np.asarray(raster_obj.data, copy=True),
```

src/nisarqa/raster_classes.py
```diff
@@ -626,34 +656,127 @@ def decimate_raster_array_to_square_pixels(
...
+
+    Returns
+    -------
+    out : numpy.ndarray
+        Copy of raster_obj.data array that has been decimated along the
+        first two dimensions to have approx. square pixels.
+    """
+    # Decimate to square pixels.
+    return decimate_array_to_square_pixels(
+        arr=raster_obj.data[...],
```
gmgunter, line 675:
```suggestion
        arr=np.asarray(raster_obj.data, copy=True),
```

src/nisarqa/input_product_readers/insar_product.py
```diff
@@ -203,5 +204,53 @@ def save_qa_metadata_to_h5(self, stats_h5: h5py.File) -> None:
                 ds_description=f"Polarizations for Frequency {freq}.",
             )
 
+    def _path_to_orbit_grp(self) -> str:
+        """Path in HDF5 to the product's orbit metadata group."""
+        pass
```
gmgunter, line 209: What's going on here? Should we delete this method?

```suggestion
```

src/nisarqa/input_product_readers/insar_product.py
```diff
@@ -203,5 +204,53 @@ def save_qa_metadata_to_h5(self, stats_h5: h5py.File) -> None:
...
+
+    def get_orbit(self, ref_or_sec: str) -> isce3.core.Orbit:
+        """
+        Get the isce3.core.Orbit for the input granule.
```
gmgunter, line 213:
```suggestion
        Get the orbit of the specified input RSLC for the input granule.
```

src/nisarqa/input_product_readers/insar_product.py
```diff
@@ -203,5 +204,53 @@ def save_qa_metadata_to_h5(self, stats_h5: h5py.File) -> None:
...
+
+        Parameters
+        ----------
+        ref_or_sec : str
+            One of {"reference", "secondary"}.
```
gmgunter, line 218:
```suggestion
        ref_or_sec : {"reference", "secondary"}
```

src/nisarqa/input_product_readers/insar_product.py
```diff
@@ -203,5 +204,53 @@ def save_qa_metadata_to_h5(self, stats_h5: h5py.File) -> None:
...
+            try:
+                orbit_grp = f[orbit_path]
+            except KeyError as e:
+                if nisarqa.Version.from_string(
+                    self.product_spec_version
+                ) < nisarqa.Version.from_string("1.1.1"):
+                    # "reference" and "secondary" orbit sub groups were
+                    # added in product specs 1.1.1
+                    orbit_path = "/".join([self._metadata_group_path, "orbit"])
+                    orbit_grp = f[orbit_path]
```
gmgunter, line 247: Is this the reference orbit? Should we raise an exception if the user asked for the "secondary" orbit?

src/nisarqa/input_product_readers/insar_product.py
```diff
@@ -203,5 +204,53 @@ def save_qa_metadata_to_h5(self, stats_h5: h5py.File) -> None:
...
+                else:
+                    raise
+            # where `group` is an `h5py.Group`
```
gmgunter, line 250:
```suggestion
```

src/nisarqa/input_product_readers/nisar_product.py
```diff
@@ -596,6 +597,65 @@ def science_freq(self) -> str:
...
+
+        Parameters
+        ----------
+        freq : str
+            Must be either "A" or "B".
+
+        Returns
+        -------
+        center_freq : float
+            The center frequency for input product's Frequency `freq`.
```
gmgunter, line 612:
```suggestion
            The center frequency for input product's Frequency `freq`, in hertz.
```

src/nisarqa/input_product_readers/nisar_product.py
```diff
@@ -596,6 +597,65 @@ def science_freq(self) -> str:
...
+        """
+        center_freq_path = f"{self.get_freq_path(freq=freq)}/centerFrequency"
```
gmgunter, line 614: This doesn't work for RSLC. It has "processedCenterFrequency" and "acquiredCenterFrequency".

https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/7b59f5f09d770952a277135b8130aac536786da9/XML/L1/nisar_L1_RSLC.xml#L294-L307

Let's move it to the InSAR product reader, rather than this base class that's common to all NISAR products.

src/nisarqa/input_product_readers/nisar_product.py
```diff
@@ -596,6 +597,65 @@ def science_freq(self) -> str:
...
+
+        Parameters
+        ----------
+        freq : str
+            Must be either "A" or "B".
+
+        Returns
+        -------
+        wavelength : float
+            The wavelength for input product's Frequency `freq`.
```
gmgunter, line 632:
```suggestion
            The wavelength for input product's Frequency `freq`, in meters.
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +492,70 @@ def truncate_colormap(
...
+                x_coordinates=x_coordinates[::arrow_stride],
+                y_coordinates=y_coordinates[::arrow_stride],
+                projection_params=quiver_projection_params,
+            )
+        )
+
+        # Invert y component because matplotlib.pyplot.quiver follows the
+        # axes convention in the background image (positive y).
+        # But, the offset images follow the top-down convention (negative y),
+        # so we need to invert the values.
```
seongsujeong, line 543:
```suggestion
        # NOTE: Invert the Y component because `matplotlib.pyplot.quiver()` after `matplotlib.pyplot.imshow()` will follow the coordinate convention of the raster i.e. row index downward. This is opposite to the coordinate system in which the offset was defined (i.e. +y upward), so the y offset needs to be inverted to properly plot the quivers on top of the image.
```

Please feel free to edit and split the lines.

src/nisarqa/input_product_readers/nisar_product.py
```diff
@@ -596,6 +597,65 @@ def science_freq(self) -> str:
...
+            The ground track velocity metadata cube.
+        """
+        grd_trk_path = "/".join(
+            [self._coordinate_grid_metadata_group_path, "groundTrackVelocity"]
+        )
+
+        with h5py.File(self.filepath, "r") as f:
+            grd_trk_vel = self._build_metadata_lut(f=f, ds_arr=f[grd_trk_path])
+
+        return grd_trk_vel
```
gmgunter, line 657:
```suggestion
            return self._build_metadata_lut(f=f, ds_arr=f[grd_trk_path])
```

seongsujeong (commented):
Nice work @nemo794. Here are my comments.

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +566,282 @@ def truncate_colormap(
...
+    azimuth and slant range directions into the X/Y directions of the projected
+    coordinates.
+
+    Parameters
+    ----------
+    epsg : int
+        EPSG of the offsets arrays and coordinate vectors.
+    orbit : isce3.core.Orbit
+        The trajectory of the radar antenna phase center over a time interval
+        that includes the observation times of each point in `geo_pts`.
```
gmgunter, line 587: What is `geo_pts`?

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +566,282 @@ def truncate_colormap(
...
+           Start of search interval, in seconds. Defaults to `orbit.start_time`.
+        'time_end':
+           End of search interval, in seconds. Defaults to `orbit.end_time`.
+        Defaults to None.
+    """
+
+    epsg: int
+    orbit: isce3.core.Orbit
+    wavelength: float
+    lookside: isce3.core.LookSide | str
```
gmgunter, line 611: Currently the docstring says "look_side", which I like better

```suggestion
    look_side: isce3.core.LookSide | str
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -55,24 +62,39 @@ def plot_offsets_quiver_plot_to_pdf(az_offset, rg_offset, params, report_pdf):
         A structure containing processing parameters to generate quiver plots.
     report_pdf : matplotlib.backends.backend_pdf.PdfPages
         The output PDF file to append the quiver plot to.
+    quiver_projection_params : None or ParamsForAzRgOffsetsToProjected, optional
+        ** Strongly Recommended for GOFF and GUNW **
+        If None, or if offsets arrays are nisarqa.RadarRaster, this will
+        be ignored. Set this to None if the contents of the offset rasters
```
gmgunter, line 68: Maybe we should raise an exception if the input is a `RadarRaster` and `quiver_projection_params is not None`?

(We probably don't want to raise an exception if the input array is a `GeoRaster` and `quiver_projection_params is None`, because that would preclude the possibility of generating QA plots for offsets that have already been re-projected.)

gmgunter (commented):
Here are notes from our abridged offline review today.

We reviewed the changes to the product readers and Raster classes, but didn't look over any of the algorithmic details yet due to time constraints.

src/nisarqa/processing/insar/quiver_plots.py
nemo794: Oh wow, this is a great explanation! Thank you!
(And thank you for figuring out the bug, too!)

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +558,282 @@ def truncate_colormap(
...
+    orbit = projection_params.orbit
+    wavelength = projection_params.wavelength
+    lookside = projection_params.lookside
+    ground_track_velocity = projection_params.ground_track_velocity
+    geo2rdr_params = projection_params.geo2rdr_params
+    geo2rdr_params = {} if (geo2rdr_params is None) else geo2rdr_params
+
+    # Useful variables for the algorithm
+    num_arrows_x_direction = len(y_coords_values_at_arrow_tails)
+    num_arrows_y_direction = len(x_coords_values_at_arrow_tails)
```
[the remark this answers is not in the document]
nemo794: Oh wow, that's a great catch. Thank you!

GOFF rasters seem to be square (usually), which means that the X and Y dimensions are the same length, so this was a silent bug 🤦‍♀️ .

gmgunter (commented):
We reviewed this together offline in our QA meeting this morning

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +568,282 @@ def truncate_colormap(
...
+    )
+
+    # Shift the radar grid's azimuth values
+    target_in_rdr_matrix[:, :, 0] += az_off_values_at_arrow_tail_in_seconds
+
+    # Shift the radar grid's range values
+    # Note: range offset values were already in meters, so just add them
+    target_in_rdr_matrix[:, :, 1] += rg_offsets_at_arrow_tails
+
+    # 4) Run geo2rdr to convert back to (x_1, y_1) LLH
```
gmgunter, line 787:
```suggestion
    # 4) Run rdr2geo to convert back to (x_1, y_1) LLH
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +568,282 @@ def truncate_colormap(
...
+            proj = isce3.core.make_projection(epsg)
+
+            # Use a dummy height value of 0 in computing the inverse projection.
+            # ISCE3 projections are always 2-D transformations -- the height
+            # has no effect on lon/lat. In ISCE3, the height valued is simply
+            # passed through and copied to the output height value, so we
+            # can ignore it.
+            lon, lat, _ = proj.inverse([x_coord, y_coord, 0])
+
+            # 2) Run rdr2geo on the (x_0, y_0) LLH
```
gmgunter, line 715:
```suggestion
            # 2) Run geo2rdr on the (x_0, y_0) LLH
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +568,282 @@ def truncate_colormap(
...
+            if not np.isfinite(aztime) or not np.isfinite(srange):
+                x_tail_coords_shifted[i, j] = np.nan
+                y_tail_coords_shifted[i, j] = np.nan
+                continue
+
+            xyz = isce3.geometry.rdr2geo_bracket(
+                # az time in seconds since orbit.reference_epoch
+                aztime=target_in_rdr_matrix[i, j, 0],
+                # slant range in meters
+                slant_range=target_in_rdr_matrix[i, j, 1],
```
gmgunter, line 810:
```suggestion
                aztime=aztime,
                # slant range in meters
                slant_range=srange,
```

You already assigned the variables `aztime` and `srange` above, so you might as well use those variables here.

https://github.com/isce-framework/nisarqa/blob/219b355eb9011e69a3fe0eb359fa1e8fdfaabba6/src/nisarqa/processing/insar/quiver_plots.py#L798-L799

gmgunter (commented):
Thanks, @nemo794! This is a really nice (and very well-documented!) improvement. Here are a few additional suggestions from a more thorough review of the implementation.

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +492,72 @@ def truncate_colormap(
...
+            )
+        )
+
+        # NOTE: Invert the Y component because `matplotlib.pyplot.quiver()`
+        # after `matplotlib.pyplot.imshow()` will follow the coordinate
+        # convention of the raster i.e. row index downward.
+        # This is opposite to the coordinate system in which the offset was
+        # defined (i.e. +y upward), so the y offset needs to be inverted
+        # to properly plot the quivers on top of the image.
+        y_offsets_at_arrow_tails *= -1
```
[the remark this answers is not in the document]
nemo794: Thank you @gmgunter for diagnosing the issue and writing a detailed assessment! Incredibly insightful. Summarizing an offline conversation:

* In this PR, we'll implement the fully-correct fix for GOFF, where we divide by the exact x and y posting
     - Note: the term "spacing" refers to the length of the side of a pixel, vs. "posting" which is a more-specific term that also includes direction. For example, a NISAR L2 product might have a y-coordinate posting of `-30`, but its y-coordinate spacing would be `30`.
* However, this new understanding illuminated a similar issue for ROFF.
    - Ideally, for correct x offsets in pixels we would divide the az offsets (meters) by (`zeroDopplerTimeSpacing` (sec) divided by ground track velocity (m/s)) to get the x offset in pixels
    - Ideally, for correct y offsets in pixels we would divide the slant range offsets (meters) by `slantRangeSpacing` (meters) to get units of pixels
* But, QA does not yet know about these metadata for ROFF, and making that change is likely beyond the scope of this PR, particularly in light of the impending Release.
* So for now, for ROFF let's use the `sceneCenterAlongTrackSpacing` and `sceneCenterGroundRangeSpacing` as approximations. (They are approximations, because the spacing will vary in different parts of the image.) The spacing should be positive for L1 products, so we shouldn't need to worry about flipping the sign on the y-axis.
* Later, we can update QA to more accurately plot the ROFF quiver arrows.

@gmgunter , please let me know if this is your understanding as well? Thank you!

src/nisarqa/processing/insar/quiver_plots.py
nemo794: d'oh! Good catch. Thanks!

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +568,282 @@ def truncate_colormap(
...
+    # Azimuth Offsets are in meters, but azimuth on the radar grid is in
+    # seconds. So, convert `az_offsets_at_arrow_tails` from meters to seconds
+    # by using the ground track velocity (meters/second) metadata cube.
+    X_tail_vals_grid, Y_tail_vals_grid = np.meshgrid(
+        x_coords_values_at_arrow_tails, y_coords_values_at_arrow_tails
+    )
+
+    # Constructing the interpolator is an expensive operation (1-2 seconds),
+    # so do this outside of the for loops.
+    grd_trk_vel_at_arrow_tail = nisarqa.interpolate_points_in_metadata_cube(
```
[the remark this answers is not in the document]
nemo794: Wow! This is great. Thank you! This should streamline and speed up the code a bit, too. 🚀

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -159,12 +272,34 @@ def plot_single_quiver_plot_to_png(
         A structure containing processing parameters to generate quiver plots.
     png_filepath : path-like
         Filename (with path) for the image PNG.
+    quiver_projection_params : None or ParamsForAzRgOffsetsToProjected, optional
+        ** Strongly Recommend providing parameters for GOFF and GUNW **
```
gmgunter, line 276:
```suggestion
        ** Strongly recommend providing parameters for GOFF and GUNW **
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -159,12 +272,34 @@ def plot_single_quiver_plot_to_png(
...
+        Set to None if the contents of the offset rasters use the same
+        coordinate grid as the raster image (e.g. both are
+        azimuth/range grid, such as for ROFF, RIFG, and RUNW).
+        If offsets arrays are nisarqa.GeoRaster and these parameters are
+        are provided, they will be used to modify the quiver arrows to
```
gmgunter, line 281:
```suggestion
        provided, they will be used to modify the quiver arrows to
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -55,23 +61,48 @@ def plot_offsets_quiver_plot_to_pdf(az_offset, rg_offset, params, report_pdf):
         A structure containing processing parameters to generate quiver plots.
     report_pdf : matplotlib.backends.backend_pdf.PdfPages
         The output PDF file to append the quiver plot to.
+    quiver_projection_params : None or ParamsForAzRgOffsetsToProjected, optional
+        ** Strongly Recommend providing parameters for GOFF and GUNW **
```
gmgunter, line 65:
```suggestion
        ** Strongly recommend providing parameters for GOFF and GUNW **
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -55,23 +61,48 @@ def plot_offsets_quiver_plot_to_pdf(az_offset, rg_offset, params, report_pdf):
...
-        raster_obj=az_offset
-    )
-    rg_off = nisarqa.decimate_raster_array_to_square_pixels(
-        raster_obj=rg_offset
-    )
+    if quiver_projection_params is not None:
+        if isinstance(az_offset, nisarqa.RadarRaster):
+            raise TypeError(
+                "Input az and rg offset rasters are instances of"
+                f" nisarqa.RadarRaster, but {quiver_projection_params=}."
```
gmgunter, line 100:
```suggestion
                f" nisarqa.RadarRaster, but {type(quiver_projection_params)=}."
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -196,6 +331,58 @@ def plot_single_quiver_plot_to_png(
...
+    # For L2 products where `quiver_projection_params` are provided,
+    # we need to adjust the geo grid parameters to match our
+    # freshly-decimated `az_off` and `rg_off`
+    kwargs = {}
+    if quiver_projection_params is not None:
+        if isinstance(az_offset, nisarqa.RadarRaster):
+            raise TypeError(
+                "Input az and rg offset rasters are instances of"
+                f" nisarqa.RadarRaster, but {type(quiver_projection_params)=}."
+                " It should not be set to None for rasters on the radar grid."
```
gmgunter, line 343:
```suggestion
                " It should be set to None for rasters on the radar grid."
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -196,6 +331,58 @@ def plot_single_quiver_plot_to_png(
...
+        x_coordinates = az_offset.x_coordinates
+        y_coordinates = az_offset.y_coordinates
+        x_coords = x_coordinates[::x_decimation]
+        y_coords = y_coordinates[::y_decimation]
+
+        assert az_off.shape[1] == len(x_coords)
+        assert az_off.shape[0] == len(y_coords)
+
+        x_posting = az_offset.x_spacing * x_decimation
+        y_posting = az_offset.y_spacing * y_decimation
```
gmgunter, line 356:
```suggestion
        x_posting = az_offset.x_posting * x_decimation
        y_posting = az_offset.y_posting * y_decimation
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -196,6 +331,58 @@ def plot_single_quiver_plot_to_png(
...
+        # For now, use `sceneCenterAlongTrackSpacing` as an approximation.
+        # (The approximation is because the spacing will vary in different
+        # parts of the image.)
+        # The spacing should be positive for L1 products, so we shouldn't
+        # need to worry about flipping the sign on the y-axis.
+        kwargs["y_posting"] = az_offset.ground_az_spacing * y_decimation
+    else:
+        kwargs["y_posting"] = az_offset.y_posting * y_decimation
+
+    kwargs["x_posting"] = az_offset.x_posting * x_decimation
```
gmgunter, line 384: Let's construct a `RadarGrid` in this case, as described in a comment above.

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -219,7 +407,13 @@ def add_magnitude_image_and_quiver_plot_to_axes(
     ax: mpl.axes.Axes,
     az_off: np.ndarray,
     rg_off: np.ndarray,
+    x_posting: float,
+    y_posting: float,
     params: nisarqa.QuiverParamGroup,
+    quiver_projection_params: (
+        None | nisarqa.ParamsForAzRgOffsetsToProjected
+    ) = None,
+    geo_grid: None | nisarqa.GeoGrid = None,
```
gmgunter, line 416:
```suggestion
    coord_grid: nisarqa.GeoGrid | nisarqa.RadarGrid,
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -219,7 +407,13 @@ def add_magnitude_image_and_quiver_plot_to_axes(
     ax: mpl.axes.Axes,
     az_off: np.ndarray,
     rg_off: np.ndarray,
+    x_posting: float,
+    y_posting: float,
```
gmgunter, line 411: Let's infer these from `coord_grid`.

```suggestion
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -239,8 +433,34 @@ def add_magnitude_image_and_quiver_plot_to_axes(
...
+        projected coordinates. Note: the plotted image (and colorbar) will
+        represent displacement magnitude in radar coordinates;
+        only the quiver arrows will be projected.
+        These parameters should correspond to `az_off` and `rg_off`.
+        Defaults to None.
+    geo_grid : None or nisarqa.GeoGrid, optional
+        GeoGrid parameters for `az_off` and `rg_off`.
+        If `quiver_projection_params` is None, this is ignored.
+        If `quiver_projection_params` is not None, this is required (it
+        cannot be None). Defaults to None.
```
gmgunter, line 463: Please update this docstring parameter name+type+description per comment above.

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -239,8 +433,34 @@ def add_magnitude_image_and_quiver_plot_to_axes(
         Along track offset raster array.
     rg_off : numpy.ndarray
         Slant range offset raster array.
+    x_posting, y_posting : float
+        Posting in the x and y directions (respectively) of `az_off`
+        and `rg_off`.
+        Note: For NISAR L2 products, the y-coordinate posting of the
+        coordinate grid is negative (the positive y-axis points up in the plot).
```
gmgunter, line 440:
```suggestion
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
     arrow_stride = int(max(np.shape(total_offset)) / params.arrow_density)
 
     # Only plot the arrows at the requested strides.
-    arrow_y = az_off[::arrow_stride, ::arrow_stride]
-    arrow_x = rg_off[::arrow_stride, ::arrow_stride]
+    az_offsets_at_arrow_tails = az_off[::arrow_stride, ::arrow_stride]
+    rg_offsets_at_arrow_tails = rg_off[::arrow_stride, ::arrow_stride]
```
gmgunter, line 584:
```suggestion
    arrow_az_offsets = az_off[::arrow_stride, ::arrow_stride]
    arrow_rg_offsets = rg_off[::arrow_stride, ::arrow_stride]
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
 
-    x = np.linspace(0, arrow_x.shape[1] - 1, arrow_x.shape[1])
-    y = np.linspace(0, arrow_x.shape[0] - 1, arrow_x.shape[0])
+    num_arrows_along_x = az_offsets_at_arrow_tails.shape[1]
+    num_arrows_along_y = az_offsets_at_arrow_tails.shape[0]
+    x = np.linspace(0, num_arrows_along_x - 1, num_arrows_along_x)
+    y = np.linspace(0, num_arrows_along_y - 1, num_arrows_along_y)
```
gmgunter, line 589:
```suggestion
    x = np.arange(num_arrows_along_x)
    y = np.arange(num_arrows_along_y)
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
+    else:
+        # GOFF, GUNW
+        # For L2 (GOFF, etc), `az_offsets_at_arrow_tails` indicates how many
+        # meters to offset in the satellite's along-track direction, which
+        # is NOT the same as the image grid's y-direction. So, do conversion.
+        if geo_grid is None:
+            raise ValueError(
+                f"`quiver_projection_params` was not None, but {geo_grid=};"
+                " it cannot be None."
+            )
```
gmgunter, line 616: We can remove this check -- with the suggested changes above, `geo_grid` becomes a required parameter (named `coord_grid`).

```suggestion
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
+
+    # According to the documentation for `matplotlib.pyplot.quiver()`,
+    # when you pass angles='xy', the arrow offsets should be in the same
+    # coordinates as the arrow tail locations, i.e. the arrows will point
+    # from (x, y) to (x+u, y+v).
+    # Here, we are storing the arrow tail locations are in pixels,
+    # while the x and y offsets are in units of meters. So, what we
+    # actually want to plot are the arrow offsets in pixels (i.e. the
+    # same units as the arrow tail locations).
+    # So let's divide the offsets by the pixel spacing.
```
gmgunter, line 641:
```suggestion
    # Arrow tail locations are in pixels. Arrow head offsets are in
    # meters. Convert from meters to pixels.
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
+    # That is, a postive-valued offset in meters will be a negative-valued
+    # offset in pixels. So, when you divide the offsets by the pixel
+    # spacing, it should flip the sign of the y component of the offsets.
+
+    # Conversely, the y-coordinate posting of NISAR L1 products is
+    # positive (the positive y-axis points down in the plot).
+    # So dividing by the pixel spacing should not flip the sign of the
+    # y component of the offsets when making quiver plots for L1 products.
+    arrow_heads_y_offsets_in_pixels = arrow_heads_y_offsets / y_posting
+    arrow_heads_x_offsets_in_pixels = arrow_heads_x_offsets / x_posting
```
gmgunter, line 654:
```suggestion
    arrow_heads_y_offsets_in_pixels = arrow_heads_y_offsets / coord_grid.y_posting
    arrow_heads_x_offsets_in_pixels = arrow_heads_x_offsets / coord_grid.x_posting
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
-        # ending y direction component of for each arrow vector
-        arrow_y * arrow_stride,
+        # starting pixel index in x direction for each arrow tail
+        arrow_tails_x_indices,
+        # starting pixel index in y direction for each arrow tail
+        arrow_tails_y_indices,
+        # x offset for each arrow's tip
+        arrow_heads_x_offsets_in_pixels * arrow_stride,
+        # y offset for each arrow's tip
+        arrow_heads_y_offsets_in_pixels * arrow_stride,
```
gmgunter, line 668: Scaling the offsets by a scalar factor won't affect the direction of the arrows -- only the magnitude. The magnitude here is basically arbitrary because Matplotlib will auto-scale them, but I don't think scaling by the stride makes intuitive sense anyway, so I'd suggest removing that. It shouldn't have any effect on the plotted output.

```suggestion
        arrow_heads_x_offsets_in_pixels,
        # y offset for each arrow's tip
        arrow_heads_y_offsets_in_pixels,
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+    Parameters
+    ----------
+    rg_offsets : numpy.ndarray
+        2D array of the values of the slant range offsets.
+        Array should be geocoded, but the pixel values should represent
+        offset in slant range direction (in meters).
+    az_offsets : numpy.ndarray
+        2D array of the values of the along track offsets.
+        Array should be geocoded, but the pixel values should represent
+        offset in along azimuth direction (in meters).
```
gmgunter, line 753: Swap the order of these in the docstring to match their order in the parameter list

```suggestion
    az_offsets : numpy.ndarray
        2D array of the values of the along track offsets.
        Array should be geocoded, but the pixel values should represent
        offset in along azimuth direction (in meters).
    rg_offsets : numpy.ndarray
        2D array of the values of the slant range offsets.
        Array should be geocoded, but the pixel values should represent
        offset in slant range direction (in meters).
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+    geo_grid : nisarqa.GeoGrid
+        GeoGrid parameters for `rg_offsets` and `az_offsets`.
```
gmgunter, line 755:
```suggestion
        Geocoded coordinate grid on which `rg_offsets` and `az_offsets` are sampled.
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+        2D array of the offset values in X and Y direction (respectively)
+        of the projected coordinate grid.
+    """
+
+    # Variable renaming and unpacking
+    rg_offsets_at_arrow_tails = rg_offsets
+    az_offsets_at_arrow_tails = az_offsets
+    epsg = geo_grid.epsg
+    x_coords_values_at_arrow_tails = geo_grid.x_coordinates
+    y_coords_values_at_arrow_tails = geo_grid.y_coordinates
```
gmgunter, line 773:
```suggestion
    arrow_tails_x = geo_grid.x_coordinates
    arrow_tails_y = geo_grid.y_coordinates
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+    Returns
+    -------
+    offset_in_x_direction, offset_in_y_direction : numpy.ndarray
+        2D array of the offset values in X and Y direction (respectively)
+        of the projected coordinate grid.
+    """
+
+    # Variable renaming and unpacking
+    rg_offsets_at_arrow_tails = rg_offsets
+    az_offsets_at_arrow_tails = az_offsets
```
gmgunter, line 770:
```suggestion
    arrow_rg_offsets = rg_offsets
    arrow_az_offsets = az_offsets
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
+
+    # Determine the offsets in the image grid's X and Y directions
+
+    if quiver_projection_params is None:
+        # ROFF, RIFG, RUNW
+        # For L1 (ROFF, etc), `az_offsets_at_arrow_tails` indicates how many
+        # meters to offset in the image grid's y-direction (aka along-track).
+        # (Similar reasoning for slant range offsets and the x-direction.)
+        arrow_heads_y_offsets = az_offsets_at_arrow_tails
+        arrow_heads_x_offsets = rg_offsets_at_arrow_tails
```
gmgunter, line 606:
```suggestion
        arrow_y_offsets = az_offsets_at_arrow_tails
        arrow_x_offsets = rg_offsets_at_arrow_tails
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+    geo2rdr_params = projection_params.geo2rdr_params
+    geo2rdr_params = {} if (geo2rdr_params is None) else geo2rdr_params
+
+    # Useful variables for the algorithm
+    num_arrows_x = len(x_coords_values_at_arrow_tails)
+    num_arrows_y = len(y_coords_values_at_arrow_tails)
+
+    if (rg_offsets_at_arrow_tails.shape[1] != num_arrows_x) or (
+        rg_offsets_at_arrow_tails.shape[0] != num_arrows_y
+    ):
```
gmgunter, line 786: What about the azimuth offsets? Do we need to check that their shape is consistent?

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+        raise ValueError(
+            f"{rg_offsets_at_arrow_tails.shape=}, but its dimensions must"
+            " correspond to the provided coordinate vectors where"
+            f" {len(geo_grid.y_coordinates)=} and"
+            f" {len(geo_grid.x_coordinates)=}/"
```
gmgunter, line 791:
```suggestion
            f" {len(geo_grid.x_coordinates)=}"
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+        )
+
+    x_tail_coords_shifted = np.zeros(
+        (num_arrows_y, num_arrows_x),
+        dtype=y_coords_values_at_arrow_tails.dtype,
+    )
+    y_tail_coords_shifted = np.zeros(
+        (num_arrows_y, num_arrows_x),
+        dtype=y_coords_values_at_arrow_tails.dtype,
+    )
```
gmgunter, line 801: Instead of this, let's just store arrays of the offsets in X and Y coordinates, which we can compute directly inside the loop below.

```suggestion
    arrow_x_offsets = np.zeros(
        (num_arrows_y, num_arrows_x),
        dtype=y_coords_values_at_arrow_tails.dtype,
    )
    arrow_y_offsets = np.zeros(
        (num_arrows_y, num_arrows_x),
        dtype=y_coords_values_at_arrow_tails.dtype,
    )
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+            # (Ideally, we'd use the target height rather than assuming zero
+            # height. But we don't have access to this information in the GOFF
+            # input product, so a reasonable approx. is to use zero height.)
+            ellipsoid = isce3.core.WGS84_ELLIPSOID
+
+            # When we used `proj.inverse()` above, we did not care
+            # about height because the height was simply passed through.
+            # Here, for `ellipsoid.lon_lat_to_xyz()` we do care about height,
+            # but we do not have a DEM in GOFF nor GUNW products.
+            # So, assume each target falls on the WGS84 ellipsoid.
```
gmgunter, line 845:
```suggestion
            # So, assume each target falls on the WGS 84 ellipsoid.
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+            # seconds by using the ground track velocity (meters/second).
+
+            # Use the aztime of each target in radar coordinates to
+            # interpolate the orbit in order to get the
+            # platform position and velocity in ECEF coordinates.
+            platform_pos, platform_vel = orbit.interpolate(aztime)
+
+            grd_trk_vel_at_arrow_tail = nisarqa.get_ground_track_velocity(
+                target_xyz,  # ECEF position of target on the ground
+                platform_pos,  # az
```
gmgunter, line 873:
```suggestion
                platform_pos,
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+            # Note: range offset values were already in meters, so just add them
+            srange += rg_offsets_at_arrow_tails[i, j]
+
+            # 4) Run rdr2geo to convert back to (x_1, y_1) LLH
+
+            # Skip NaN pixels (this is usually geocoding fill)
+            if not np.isfinite(aztime) or not np.isfinite(srange):
+                x_tail_coords_shifted[i, j] = np.nan
+                y_tail_coords_shifted[i, j] = np.nan
+                continue
```
gmgunter, line 894: This should happen immediately after computing `aztime` and `srange` using geo2rdr above. Let's move it up there.

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+            # Shift the radar grid's range values
+            # Note: range offset values were already in meters, so just add them
+            srange += rg_offsets_at_arrow_tails[i, j]
+
+            # 4) Run rdr2geo to convert back to (x_1, y_1) LLH
+
+            # Skip NaN pixels (this is usually geocoding fill)
+            if not np.isfinite(aztime) or not np.isfinite(srange):
+                x_tail_coords_shifted[i, j] = np.nan
+                y_tail_coords_shifted[i, j] = np.nan
```
gmgunter, line 893:
```suggestion
                arrow_x_offsets[i, j] = np.nan
                arrow_y_offsets[i, j] = np.nan
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+                # of the image grid, use 0.0.
+                doppler=0.0,
+                # Technically the wavelength is not needed here since
+                # Doppler is zero and the wavelength is just multiplied
+                # by the Doppler. But since we already need the wavelength
+                # for geo2rdr, might as well use it.
+                wavelength=wavelength,
+                # Ideally we would use the actual DEM, but we don't have
+                # this information available so use a zero-height DEM.
+                dem=isce3.geometry.DEMInterpolator(),
```
gmgunter, line 916:
```suggestion
                # Pass `epsg=4326` to ensure the DEM is referenced to the
                # WGS 84 ellipsoid.
                dem=isce3.geometry.DEMInterpolator(epsg=4326),
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+
+            # Once again, ignore the returned height
+            lon, lat, _ = ellipsoid.xyz_to_lon_lat(xyz)
+
+            # 5) Convert shifted LLH (x_1, y_1) back into projected
+            #    coordinates
+            # To convert from LLH back to the projected coordinates of the
+            # input arrays, use the forward() method of the same `proj`
+            # object we created above.
+            xyz = proj.forward([lon, lat, 0])
```
gmgunter, line 927: Let's not call this `xyz`. We use the variable name `xyz` above to represent the location of the shifted target in ECEF coordinates. Now we're using it to represent the location in projected coordinates.

```suggestion
            x_shift, y_shift, _ = proj.forward([lon, lat, 0])
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+            x_tail_coords_shifted[i, j] = xyz[0]
+            y_tail_coords_shifted[i, j] = xyz[1]
```
gmgunter, line 929:
```suggestion
            # 6) Compute new offset values in projected coordinates:
            #       (x_1 - x_0, y_1 - y_0)
            arrow_x_offsets[i, j] = x_shift - x_coord
            arrow_y_offsets[i, j] = y_shift - y_coord
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+
+    # 6) Compute new offset values in projected coordinates:
+    #       (x_1 - x_0, y_1 - y_0)
+    X_tail_vals_grid, Y_tail_vals_grid = np.meshgrid(
+        x_coords_values_at_arrow_tails, y_coords_values_at_arrow_tails
+    )
+
+    offset_in_x_direction = x_tail_coords_shifted - X_tail_vals_grid
+    offset_in_y_direction = y_tail_coords_shifted - Y_tail_vals_grid
```
gmgunter, line 938:
```suggestion
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -382,4 +683,261 @@ def truncate_colormap(
...
+
+    return offset_in_x_direction, offset_in_y_direction
```
gmgunter, line 940:
```suggestion
    return arrow_x_offsets, arrow_y_offsets
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -352,26 +580,99 @@ def truncate_colormap(
...
+        # For L2 (GOFF, etc), `az_offsets_at_arrow_tails` indicates how many
+        # meters to offset in the satellite's along-track direction, which
+        # is NOT the same as the image grid's y-direction. So, do conversion.
+        if geo_grid is None:
+            raise ValueError(
+                f"`quiver_projection_params` was not None, but {geo_grid=};"
+                " it cannot be None."
+            )
+
+        arrow_heads_x_offsets, arrow_heads_y_offsets = (
```
gmgunter, line 618:
```suggestion
        arrow_x_offsets, arrow_y_offsets = (
```

src/nisarqa/raster_classes.py
```diff
@@ -414,7 +581,100 @@ def x_axis_label(self) -> str:
...
+
+    The attributes specified here are based on the needs of the QA code
+    for generating and labeling plots, etc.
+
+    Parameters
+    ----------
+    epsg : int
+        The EPSG code of the input raster.
+    x_spacing : float
+        X posting of pixels (in meters) of input array.
```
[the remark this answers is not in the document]
nemo794: Updated to: "X posting of pixels of input array, in units matching `x_coordinates`."

src/nisarqa/raster_classes.py
```diff
@@ -414,7 +581,100 @@ def x_axis_label(self) -> str:
...
+    x_coordinates : numpy.ndarray
+        1D vector of the coordinate values of the center of each pixel
+        of the raster grid in the X direction.
```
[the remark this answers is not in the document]
nemo794: Updated ending to "...the raster grid in the X direction, in the units of `epsg`."

src/nisarqa/raster_classes.py
```diff
@@ -444,36 +718,30 @@ class GeoRaster(SARRaster):
...
-    y_stop: float
+    Notes
+    -----
+    All initialization parameters will also be stored as attributes.
+    """
 
-    epsg: int
+    def __post_init__(self):
+        # Initialize the start and stop attributes
+        super().__post_init__()
```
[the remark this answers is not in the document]
nemo794: It looks like `super().__post_init__()` follows MRO order, until it finds a defined `__post_init__()`: https://stackoverflow.com/a/3277407

It seems like it'd be a good thing to let Python do its thing; the trick is that we need to ensure that `GeoGrid`'s `__post_init__` gets called at some point.

What about something like this ?

```python
class GeoRaster(SARRaster, GeoGrid):
    """"""
    def __post_init__(self):
        # Make sure to initialize the start and stop attributes
        GeoGrid.__post_init__(self)

        # Now, follow MRO for standard __post_init__
        super().__post_init__()
```

src/nisarqa/raster_classes.py
```diff
@@ -449,32 +449,65 @@ class GeoRaster(SARRaster):
...
+    y_stop: float = field(init=False)
 
     epsg: int
 
+    x_coordinates: np.ndarray
+    y_coordinates: np.ndarray
+
+    def __post_init__(self):
+        # Infer start and stop values
+        self.x_start = float(self.x_coordinates[0])
```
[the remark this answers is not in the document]
nemo794: Update: Resolved in this PR via:
https://github.com/isce-framework/nisarqa/pull/67#discussion_r2323512642
https://github.com/isce-framework/nisarqa/pull/67#discussion_r2323512972
https://github.com/isce-framework/nisarqa/pull/67#discussion_r2323514372
https://github.com/isce-framework/nisarqa/pull/67#discussion_r2323516809

nemo794:
Phew! All comments addressed. Incredibly helpful and insightful comments. Great catches on bugs, too! 🚀 

Remaining: One set of comments has an open question. And, I need to raise a couple of Issues to address later. :)

src/nisarqa/raster_classes.py
```diff
@@ -414,7 +581,100 @@ def x_axis_label(self) -> str:
...
+
+        # Y in meters (units are specified as meters in the product spec)
+        # For NISAR, geocoded grids are referenced by the upper-left corner
+        # of the pixel to match GDAL conventions. So add the distance of
+        # the pixel's side to bottom to get the actual stop value.
+        self.y_stop = float(self.y_coordinates[-1] + self.y_spacing)
+
+    @property
+    def x_posting(self):
+        return self.x_spacing
```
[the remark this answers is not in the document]
nemo794: Please see Issue #72

PR #70, opened 2025-08-29
Add sanity check for `staticLayersDataAccess`

nemo794:
The `staticLayersDataAccess` metadata for L2 products was moved to the `identification` Group in product specs v1.4.0. This PR reflects that change.

### DIFF
--- a/src/nisarqa/validate/sanity_checks.py
+++ b/src/nisarqa/validate/sanity_checks.py
@@ -435,6 +435,9 @@ def identification_sanity_checks(
             "plannedObservationId",
         ]
 
+    if product_type.upper().startswith("G"):
+        misc_ds += ["staticLayersDataAccess"]
+
     for ds_name in misc_ds:
         ds_checked.add(ds_name)
         if _dataset_exists(ds_name):


src/nisarqa/validate/sanity_checks.py
```diff
@@ -435,6 +435,9 @@ def _verify_data_is_in_list(
             "plannedObservationId",
         ]
 
+    if product_type.upper().startswith("G"):
```
gmgunter, line 438: Haha we need some helper functions like

```python
def is_geocoded_product(product_type: str) -> bool:
    ...
```

and

```python
def is_insar_product(product_type: str) -> bool:
    ...
```

gmgunter (approved):
LGTM, nice catch!

src/nisarqa/validate/sanity_checks.py
gmgunter: Oh sorry, it probably wasn't clear -- please don't implement these helper functions in this PR :pray:. It's good-to-go as is.

nemo794: Already raised as an Issue #71 ;)

nemo794:
Thank you, @gmgunter !

PR #87, opened 2025-09-16
(Non-InSAR) Add option to generate individual PNG+KMLs for each imagery layer

nemo794:
This PR adds the feature for RSLC, GSLC, and GCOV QA to output a grayscale PNG+KML for each individual imagery layer, in addition to the primary browse image PNG+KML.

Each additional PNG+KML filename is suffixed with the frequency and polarization/term of the image. For example:
* RSLC/GSLC Freq A, Pol HV would lead to these files: `BROWSE_A_HV.png` and `BROWSE_A_HV.kml`
* GCOV Freq B, Term HVVV would lead to these files: `BROWSE_B_HVVV.png` and `BROWSE_B_HVVV.kml`

By default, this is set to "False" so that we do not change the interface with SDS and ASF.

cc: @hfattahi


gmgunter (approved):
LGTM, thanks!

PR #90, opened 2025-09-22
Remove HSI feature from RIFG/RUNW/GUNW

### LINKED ISSUE #78: Remove HSI plot from RIFG/RUNW/GUNW Report PDF
Currently, QA for RIFG/RUNW/GUNW generates an "HSI" image in the report PDF. For now, let's remove this plot. (details below)

## Background

* Scientists often use Hue-Saturation-Intensity (HSI) images for analysis of interferometric products. These are generated by combing the interferogram layer ("hue") with complex data ("intensity") from the input RSLCs.

* Because these images can combine two pieces of information into a single image, the original plan for NISAR was to use an HSI image for the RIFG/RUNW/GUNW browse images.

* By design, `nisarqa` only knows about the the input RIFG/RUNW/GUNW granule; QA does not know about the source RSLCs (aka the SLC data) which are inputs to the InSAR products.
    - This means that when the "HSI" image was implemented in QA, it uses:
        - the phase layer for "hue"
        - all `1`s for "saturation"
        - the coherence magnitude layer for "intensity" -- **not** the RSLC's complex data
    - QA "HSI" Implementation here: https://github.com/isce-framework/nisarqa/blob/5f0434f28537662d310576663699ac1249f2d81d/src/nisarqa/processing/insar/hsi.py#L250
    - QA's "HSI" image was 1) used to generate the browse PNG, and 2) appended to the Report PDF with a colorbar.

* After implementation, the science team raised concerns about the use of the coherence magnitude layer instead of the complex data for "intensity". Additionally, areas of very low coherence would appear mostly black and thus "hide" the phase information in HSI images; this would be unhelpful/unsightly for a browse image, particularly in cases where a majority of a frame had low coherence.
    - Because of these concerns, the plan changed and it was decided to **not** use the "HSI" image as the browse image.
   - QA was updated to use the interferogram (phase) image only for the RIFG/RUNW/GUNW browse images.

Furthermore, because a ready-made HSI-to-RGB conversion function was [not available at the time of initial implementation](https://github.com/isce-framework/nisarqa/blob/5f0434f28537662d310576663699ac1249f2d81d/src/nisarqa/processing/insar/hsi.py#L266-L270), the current QA implementation uses the `matplotlib.hsv_to_rgb()`. There is a mathematical difference between HSI and HSV, and @vbrancat confirmed that HSI is definitely the correct color model.

## Current request

Because of the two issues mentioned above (use of the coherence magnitude layer, and use of HSV-to-RGB conversion function), and because the "HSI" image is no longer used for the browse image PNGs, the QA team recommends removing the "HSI" plot from the Report PDF.

## Future work

After discussion with @vbrancat , there is thought that the necessary data layer derived from the source RSLCs might be added into the RIFG/RUNW/GUNW products, which would allow the correction formation of an HSI image. At that time, QA should reconsider adding back in the HSI plot, but this time with the preferred data for "intensity", and with the correct HSI-to-RGB implementation.

## Current requests

@vbrancat , Could you please:
* Confirm the description above matches your understanding of the Issue, and update/correct/add details as needed?
* Confirm if QA should (or should not) remove the current "HSI" plot, due to the above-mentioned technical concerns?
* In case it is decided to re-implement the HSI image in the future, (but this time in the correct way!), could you please provide details on what the correct "intensity" layer should be?
    - i.e. either how to derive it from the RSLC, and/or what it would be called if/when added into RIFG/RUNW/GUNW products

Thank you!

Example of the current "HSI" plot for GUNW:

Image

nemo794:
Closes #78 .


nemo794:
Thank you, @gmgunter !

I rebased this PR branch, and then addressed the comment. (Plus, a couple of minor docstring cleanups I noticed.)

Please let me know if there's anything else! Thank you!

nemo794:
Thank you, @gmgunter !

FYI - chatted offline with @hfattahi , who agreed that it's okay to merge this PR and remove the HSI images from the PDFs.

PR #142, opened 2025-10-01
Bugfix: Antimeridian crossing

nemo794:
### Issue

#27  introduced a bug into QA where certain frames over the antimeridian had their longitude values unwrapped incorrectly for the KML, resulting in outputs which plotted like this:

Screenshot 2025-10-01 at 2 41 12 PM

Looking at the longitude values in the KML screenshot, the unwrapping issue becomes obvious.

### Bugfix // Strategy

Original code snippet with bug, where the original `prev_lon` was being used as the reference for determining the `next_lon`, instead of using the updated `prev_lon`:

```python
    unwrapped = [lons_180[0]]
    for prev_lon, curr_lon in nisarqa.pairwise(lons_180):
        delta = curr_lon - prev_lon
        if delta > 180:
            curr_lon -= 360
        elif delta < -180:
            curr_lon += 360
        unwrapped.append(curr_lon)
    return unwrapped
```

Simple correction for that code snippet:

```python
    unwrapped = [lons_180[0]]
    prev_lon = lons_180[0]
    for curr_lon in lons_180[1:]:
        delta = curr_lon - prev_lon
        if delta > 180:
            curr_lon -= 360
        elif delta < -180:
            curr_lon += 360
        unwrapped.append(curr_lon)
        prev_lon = curr_lon
    return unwrapped
```

However, making this simple change does not account for if/when a polygon wraps around the globe multiple times.

So, @gmgunter suggested using an implementation similar to ISCE3's [`unwrap_degrees()`](https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/python/packages/isce3/geometry/bounding_polygon.py#L46-L67), which leverages NumPy's built-in unwrapping to correctly handle all cases.

This PR implements that change, resulting KMLs with correctly-unwrapped longitudes which plot nicely:

Screenshot 2025-10-01 at 2 57 59 PM

### DIFF
--- a/src/nisarqa/lonlat.py
+++ b/src/nisarqa/lonlat.py
@@ -2,12 +2,12 @@ from __future__ import annotations
 
 import os
 import textwrap
-from collections.abc import Generator, Iterable, Sequence
+from collections.abc import Iterable, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any
 
 import numpy as np
+from numpy.typing import ArrayLike
 
 import nisarqa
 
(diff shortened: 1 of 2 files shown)


src/nisarqa/lonlat.py
```diff
@@ -37,26 +37,34 @@ def unwrap_longitudes(lons: Iterable[float]) -> list[float]:
...
 
-    return unwrapped
+    Returns
+    -------
+    numpy.ndarray
+        The unwrapped angles, in degrees.
+    """
+    # `numpy.unwrap` doesn't correctly unwrap inputs in degrees prior
+    # to NumPy 1.21.0. See https://github.com/numpy/numpy/pull/16987.
+    # ISCE3 currently supports NumPy >=1.20.
```
gmgunter:
```suggestion
    # nisarqa currently supports NumPy >=1.20.
```

gmgunter (approved):
LGTM, thanks for the fix!

nemo794:
Thank you, @gmgunter ! 🚀

PR #147, opened 2025-10-29
Turn off caching for GSLC QA

nemo794:
Benchmarking after the R5 release showed a slowdown in GSLC QA for full-size datasets.

Further testing reproduced this slowdown, and confirmed that it was caused by the use of the cache. The testing also showed that for full-size GSLC dataset, when each layer was decompressed and stored as a memory-mapped temp file, each file is signficantly larger than the source GSLC HDF5. For example, a 20+5 DHDH GSLC HDF5 is 13GB but it decompressed to two ~20GB mem-mapped files for FreqA, two ~5GB mem-mapped files FreqB (~50GB total) on the scratch disk. 40+5 datasets would decompress to ~40GB per FreqA layer, etc.

Our sense is that the cost of re-reading and moving that large uncompressed data volume for each access by QA was greater than the time cost of re-reading and moving the smaller compressed layer and paying the decompression costs each time.

This PR turns off the caching for GSLC by default to improve runtime during operations, until we're able to investigate further and perhaps find a more-elegant solution.

### DIFF
--- a/src/nisarqa/parameters/gslc_params.py
+++ b/src/nisarqa/parameters/gslc_params.py
@@ -6,9 +6,9 @@ import nisarqa
 from .caltools_params import PointTargetAnalyzerParamGroup
 from .nisar_params import (
     InputFileGroupParamGroup,
+    NoCacheSoftwareConfigParamGroup,
     ProductPathGroupParamGroup,
     RootParamGroup,
-    SoftwareConfigParamGroup,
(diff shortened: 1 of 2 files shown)


Tyler-g-hudson (approved):
LGTM!

nemo794:
Thanks, @Tyler-g-hudson !

PR #151, opened 2026-01-22
Update `CUSTOM_CYCLER` to have 10 distinct line styles for plotting

### LINKED ISSUE #150: Update `CUSTOM_CYCLER` to have 10 line styles
### Issue

Currently, QA's [`CUSTOM_CYLER`](https://github.com/isce-framework/nisarqa/blob/9dc63cac4d18d25e3d74c6c5036a85fcbf2ea02e/src/nisarqa/globals.py#L23-L36) repeats line colors+line styles if there are more than 6 lines plotted.

As noted by @gshiroma , all GCOV QP data will be produced full covariance, so in e.g. the histogram plots there will 10 layers plotted and no way to differentiate certain layers. For example, in this plot, there's no way to differentiate HHHH from HHVV, or HVHV from HVVH:

Image

### Fix

QA's QA's [`CUSTOM_CYLER`](https://github.com/isce-framework/nisarqa/blob/9dc63cac4d18d25e3d74c6c5036a85fcbf2ea02e/src/nisarqa/globals.py#L23-L36) needs to be updated to support 10 distinct line styles.

nemo794:
This closes #150 

This update to `nisarqa`'s `CUSTOM_CYCLER` global impacts the QA PDF plots which use line styles for all eight L1/L2 products, although GCOV QA PDF reports are the only which need all 10 distinct line styles.

As an example output from this update, for quad pol GCOV products, the Freq A backscatter histograms will now be plotted with 10 distinct line styles like this:

Screenshot 2026-01-22 at 2 11 12 PM

### DIFF
--- a/src/nisarqa/globals.py
+++ b/src/nisarqa/globals.py
@@ -21,20 +21,24 @@ SEABORN_COLORBLIND = [
     (0.33725490196078434, 0.7058823529411765, 0.9137254901960784),
 ]
 CUSTOM_CYCLER = (
-    cycler(color=SEABORN_COLORBLIND[:6])
+    cycler(color=SEABORN_COLORBLIND[:10])
+    # repeat/mix line styles and widths to reach 10 unique combinations
     + cycler(
         linestyle=[
             "-",
             "-.",
             "--",
+            ":",
             (0, (3, 1, 1, 1)),
             (0, (3, 5, 1, 5, 1, 5)),
+            "-",
+            "-.",
+            "--",
             ":",
         ]
     )
-    + cycler(lw=np.linspace(3, 1, 6))
+    + cycler(lw=[1.5, 2.0, 2.5, 3.0, 1.5, 2.0, 2.5, 3.0, 2.0, 1.5])
 )
-
 
 #: The UTC time at which this module was first loaded, in ISO 8601 format with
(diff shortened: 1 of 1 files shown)


gshiroma:
Thank you, @nemo794 , for issuing this PR. The new color palette solves the problem, but I don't think it looks good. There's too much going on, maybe too many colors. One way to improve this might be to use fewer colors and introduce different line styles instead. For example, you could keep the first four colors and use three line patterns, which would give you twelve distinct combinations before they start repeating. You can test some different combinations.

I’d suggest moving the green color to be second in the sequence. The HV polarization is usually the second element, and cross-polarimetric terms are often associated with vegetation. This convention would apply to both RSLC and GSLC products as well.

Also, is there a way to emphasize the diagonal terms? It looks like they’re plotted first and then get buried beneath the other curves. At the same time, it's nice that they appear first in the legend. Some ways to fix it:
- Use `[zorder](https://matplotlib.org/3.1.1/gallery/misc/zorder_demo.html)`.
- Change the plotting order, and keep the legend order by setting it manually (e.g., using `handles` in the call to `plt.legend()`.
- Increase the linewidth for the first elements in the cycler (or decrease it for the later ones).

Thanks again.

PR #160, opened 2026-03-10
Bugfix: Handle NFS silly rename issue when cleaning up scratch directory   

### LINKED ISSUE #152: `nisarqa` does not exit cleanly on certain filesystems (NFS, NISAR On-Demand)
When `nisarqa` uses local filesystems for the scratch directory (such as on a local computer, or in mission operations), it exits successfully.

However, when `nisarqa` uses Network File Systems (NFS) or the NISAR On-Demand system, it successfully generates all outputs but then fails to exit cleanly, raising the Exception `"QA SAS program failed with exit code 1"` when calling `shutil.rmtree()` [here](https://github.com/isce-framework/nisarqa/blob/9dc63cac4d18d25e3d74c6c5036a85fcbf2ea02e/src/nisarqa/utils/utils.py#L761) to cleanup the scratch directory.

In these cases, the error appears similar to this:

```
Starting Quality Assurance for input file: output/rslc.h5
Input file validation complete.
`qa_reports` processing complete.
Absolute Radiometric Calibration CalTool complete.
Noise Equivalent Backscatter CalTool complete.
Point Target Analyzer CalTool complete.
QA SAS complete. For details, warnings, and errors see output log file.
Traceback (most recent call last):
  File "/opt/conda/bin/nisarqa", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/nisarqa/__main__.py", line 219, in main
    run()
  File "/opt/conda/lib/python3.12/site-packages/nisarqa/__main__.py", line 190, in run
    with nisarqa.create_unique_subdirectory(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/opt/conda/lib/python3.12/site-packages/nisarqa/utils/utils.py", line 761, in create_unique_subdirectory
    shutil.rmtree(path)
  File "/opt/conda/lib/python3.12/shutil.py", line 759, in rmtree
    _rmtree_safe_fd(stack, onexc)
  File "/opt/conda/lib/python3.12/shutil.py", line 703, in _rmtree_safe_fd
    onexc(func, path, err)
  File "/opt/conda/lib/python3.12/shutil.py", line 662, in _rmtree_safe_fd
    os.rmdir(name, dir_fd=dirfd)
OSError: [Errno 39] Directory not empty: PosixPath('/nobackupnfs1/nisar/mcayanan/track_frame_002_117_056_partial_individual_L_05_SV_00_NA_23_state-config/scratch/qa-rslc-20260115152651-tg_byu2_')
2026-01-15T15:27:21.093509Z, Error, PGE::rslc_l_pge.py, run_sas_program, 303004, /opt/pge/l1_l2/sas_runner.py:79, "QA SAS program failed with exit code 1"
```

As described by M Cayaman, here is what's happening:

> The reason [for this] specific error is due to a well-known behavior in Network File Systems (NFS) called "NFS Silly Rename."
> 
> It is essentially a "race condition" between your software trying to delete a folder and the network storage system confirming that the files inside are actually gone.
> 
> Here is the step-by-step breakdown of why it failed:
> 
> 1. The Trigger
> The nisarqa Python script uses a context manager (the with statement) to handle its scratch space.
>    - Start: It creates a unique folder: .../scratch/qa-rslc-....
>    - Work: It writes temporary files into that folder and reads them back.
>    - Finish: As soon as the job finishes, the script instantly runs shutil.rmtree() to delete the folder and everything inside it.
> 
> 2. The Conflict (The "Race")
> On a local disk (like the one on your laptop or the NVMe drive on AWS), file deletion is instant.
> On NFS (/nobackupnfs1), it works differently:
>    - When shutil.rmtree() deletes a file, if any process (even the kernel cache) still has a "handle" open on that file for even a microsecond, NFS cannot delete it immediately.
>    - Instead, NFS renames the file to a hidden temporary name like .nfs0000001234. This preserves the data until the handle is fully closed. This is the "Silly Rename."
> 
> 3. The Crash
> The cleanup script runs effectively instantly:
>    - Step A: Delete file temp_data.dat.
>         * NFS Response: "Wait, a handle is lingering. I'll rename it to .nfs123 for a millisecond while I clean up."
>    - Step B: Delete directory qa-rslc-....
>         * System Check: "Is the directory empty?"
>         * Reality: No. It contains the hidden file .nfs123.
>    - Result: The command fails with OSError: [Errno 39] Directory not empty.
>

nemo794:
## Summary                                                                                                                  
  Fixes #152                                                                                                                  
                                                                                                                              
  This PR resolves the issue where `nisarqa` fails to exit cleanly when using Network File Systems (NFS) or the NISAR         
  On-Demand system (when the scratch directory is located in `/data`, which uses nfs4) due to the "NFS silly rename" behavior.                                                                    
                                                                                                                              
  ## Changes                                                                                                                  
  - Updated `create_unique_subdirectory()` context manager in `src/nisarqa/utils/utils.py` to use an `onerror` callback with `shutil.rmtree()`                                                                                                           
  - The callback gracefully handles:                                                                                          
    - `FileNotFoundError` : Same handling as the current implementation, which alert developers if fresh code violates the `nisarqa` design pattern by accidentally deleting the scratch directory outside of its context manager
    - `OSError` with `ENOTEMPTY`/`EEXIST`: Logs a warning but doesn't raise, allowing clean exit on NFS where files are temporarily renamed to `.nfsXXXXX` before deletion                                                                          
    - Other errors: Still raises as expected                                                                                  
                                                                                                                              
  ## Technical Details                                                                                                        
  The "NFS silly rename" behavior occurs when a file is unlinked while still open. NFS renames it to `.nfsXXXXX` temporarily instead of deleting immediately, which can cause "Directory not empty" errors during cleanup.                               
                                                                                                                              
  This implementation follows the pattern used in CPython's `tempfile.TemporaryDirectory` and uses `onerror` (rather than the newer `onexc` introduced in Python 3.12) for compatibility with `isce3` and `nisarqa`'s requirement of Python 3.9+     
                                                                                                                              
  ## Testing                                                                                                                  
  - The fix allows `nisarqa` to exit cleanly on NFS and NISAR On-Demand systems                                               
  - Local filesystem behavior remains unchanged                                                                               
  - Compatible with Python 3.9+

### DIFF
--- a/src/nisarqa/utils/utils.py
+++ b/src/nisarqa/utils/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import errno
 import logging
 import os
 import shutil
@@ -755,10 +756,46 @@ def create_unique_subdirectory(
     try:
         yield path
     finally:
-        log = nisarqa.get_logger()
         if delete:
+            log = nisarqa.get_logger()
+            # Use onerror callback to handle NFS "silly rename" behavior
+            # See: https://github.com/isce-framework/nisarqa/issues/152
+
+            # shutil.rmtree's onerror was deprecated in Python 3.12 and was
+            # planned to be removed in 3.14, but its deprecation was changed
+            # to docs only, with no pending removal version.
+            # See: https://github.com/python/cpython/pull/118947
+            def onerror(func, error_path, exc_info):
+                """Error handler for shutil.rmtree to handle NFS issues."""
+                exc_type, exc_value, exc_tb = exc_info
+
+                if issubclass(exc_type, FileNotFoundError):
+                    # File was already deleted (possibly by NFS), ignore
+                    pass
+                elif issubclass(exc_type, OSError):
+                    # Check if this is a "Directory not empty" error
+                    # This can occur on NFS due to "silly rename" (.nfsXXXXX files)
+                    if exc_value.errno in (errno.ENOTEMPTY, errno.EEXIST):
+                        # Log but don't raise - allow clean exit on NFS
+                        log.warning(
+                            f"nisarqa Could not delete '{error_path}': "
+                            f" {exc_value}. This may occur on Network File"
+                            " Systems where files are temporarily renamed"
+                            " before deletion."
+                        )
+                    else:
+                        # Some other OSError - raise it
+                        log.error(f"Error deleting '{error_path}': {exc_value}")
+                        raise
+                else:
+                    # Unexpected error type - raise it
+                    log.error(
+                        f"Unexpected error deleting '{error_path}': {exc_value}"
+                    )
+                    raise
+
             try:
-                shutil.rmtree(path)
+                shutil.rmtree(path, onerror=onerror)
             except FileNotFoundError:
                 msg = (
                     f"Created directory was '{path}', but it was"
--- a/src/nisarqa/utils/utils.py
+++ b/src/nisarqa/utils/utils.py
@@ -770,11 +770,20 @@ def create_unique_subdirectory(
                 exc_type, exc_value, exc_tb = exc_info
 
                 if issubclass(exc_type, FileNotFoundError):
-                    # File was already deleted (possibly by NFS), ignore
-                    pass
-                elif issubclass(exc_type, OSError):
-                    # Check if this is a "Directory not empty" error
-                    # This can occur on NFS due to "silly rename" (.nfsXXXXX files)
+                    # Directory was already deleted. Notify users because likely
+                    # they are violating a nisarqa design pattern; they
+                    # should fix their code to prevent tricky runtime bugs.
+                    msg = (
+                        f"Created directory was '{path}', but it was"
+                        " deleted external to (but within the context of)"
+                        " this context manager."
+                    )
+                    log.error(msg)
+                    raise FileNotFoundError(msg)
+
+                if issubclass(exc_type, OSError):
+                    # Check if this is a "Directory not empty" error. This can
+                    # occur on NFS due to "silly rename" (.nfsXXXXX files)
                     if exc_value.errno in (errno.ENOTEMPTY, errno.EEXIST):
                         # Log but don't raise - allow clean exit on NFS
                         log.warning(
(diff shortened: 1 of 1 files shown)


nemo794:
cc: @mcayanan

rad-eng-59:
Thanks @nemo794  for the fix. Just tested your branch w/ latest ISCE3 develop branch on ADT ondemand and all went well. The final message on screen was: 
```
QA SAS complete. For details, warnings, and errors see output log file.
```
No cleanup error this time! I wanted to approve your PR but apparently I am not allowed!
Btw, who is this @gmgunter ? Seems like a powerful guy with some influence and people inside our organization :stuck_out_tongue_winking_eye:

bhawkins:
I can confirm this fixed my issue on the NISAR on-demand system. Thanks!

bhawkins:
Well, looking closer it seems like nisarqa didn't fail with an error message, but it did leave behind an empty directory in the scratch location.

bhawkins:
Looking even more closely, I see in the log file (not stdout/stderr) a message about failing to delete the directory. So I guess this is working as intended. (Pardon the spam.)

rad-eng-59 (approved):
LGTM. Thanks.

mcayanan (approved):
LGTM

src/nisarqa/utils/utils.py
```diff
@@ -755,20 +756,56 @@ def create_unique_subdirectory(
...
-                log.info(f"Directory deleted recursively: '{path}'")
+            log = nisarqa.get_logger()
+            # Use onerror callback to handle NFS "silly rename" behavior
+            # See: https://github.com/isce-framework/nisarqa/issues/152
+
+            # shutil.rmtree's onerror was deprecated in Python 3.12 and was
+            # planned to be removed in 3.14, but its deprecation was changed
+            # to docs only, with no pending removal version.
+            # See: https://github.com/python/cpython/pull/118947
+            def onerror(func, error_path, exc_info):
```
Tyler-g-hudson, line 768: The `onerror` pattern in `shutil.rmtree` has been replaced by `onexc` - `onexc` functions receive only an exception parameter. While I also haven't found documentation anywhere of intent to remove `onerr`, it may still be more future-proof to switch to `onexc`

Tyler-g-hudson (approved):
LGTM generally! I have one thought regarding the `onerror` parameter to `shutil.rmtree`, but in general things appear to be working!

src/nisarqa/utils/utils.py
nemo794: Ah, good idea to be looking ahead!

As noted in the PR description, `onexc` was introduced in Python 3.12. Since `isce3` (and by extension, `nisarqa`) need to be backwards compatible to Python 3.9 (or is it Python 3.10 now? either way..), we cannot use `onexc` here.

Fortunately, it does not look like they're going to remove `onerror` for the time being (See: https://github.com/python/cpython/pull/118947), so we should be good for awhile.

In the future if/when `onerror` does get fully removed, then the nisarqa developer will need to reassess the best approach and update this code.

nemo794:
> Looking even more closely, I see in the log file (not stdout/stderr) a message about failing to delete the directory. So I guess this is working as intended. (Pardon the spam.)

Hmm... @bhawkins would a preferred behavior be to also emit this message to stdout/stderr? So that way a user would be reminded to periodically cleanup the empty directories from the scratch directory?

PR #161, opened 2026-04-30
(All L1/L2 Products) Improved Geolocation for Browse PNGs and KMLs

### LINKED ISSUE #61: KML Images from NISAR Products May Appear Distorted
The browse image of a NISAR product is generated in the product’s native coordinates (UTM or polar stereographic). The KML file is created from the browse image using the location of its four corners converted to latitude and longitude. Consequently, the KML images may appear distorted when viewed in tools such as Google Earth, particularly for products covering regions near the poles.

Example of a KML image generated with `nisarqa` opened in Google Earth from a GCOV product (REE) simulated over Antarctica (latitude –69 deg):

Image

The same GCOV product displayed from a KMZ file generated with [PLAnT](https://gitlab.com/plant/plant):

Image

nemo794:
This PR closes #61 .

  # Improve Browse Image Geolocation Accuracy and Add EPSG 4326 Support

  ## Summary

  This PR resolved two critical geolocation issues with NISAR QA browse images and adds optional EPSG 4326 (lat/lon) browse generation for all 8 product types (RSLC, GSLC, GCOV, RIFG, RUNW, GUNW, ROFF, GOFF).

  ## Problems Solved

  ### Issue 1: Incorrect Coordinate Grid for Browse Images
  **Problem:** Browse PNG images retained their source coordinate grid (range-doppler for L1, UTM/polar stereo for L2),
  making the image's pixels not correctly geolocated in standard GIS tools expecting lat/lon coordinates.

  **Solution:** Added optional generation of "EPSG 4326 (lat/lon) Browse" images alongside the original "Primary Browse", enabled via `output_browse_4326: true` in runconfig.

  ### Issue 2: Inaccurate KML Corner Coordinates
  **Problem:** KML corner coordinates were computed from the full-resolution source raster but browse images are
  multilooked/decimated to smaller sizes. This caused browse images to not align correctly when overlaid in Google Earth.

  **Solution:** Now compute KML corners from the actual decimated/multilooked coordinate grids, ensuring pixel-perfect
  alignment.

  ## Impact

  ### User Benefits
  - Optional lat/lon browse images now overlay accurately in Google Earth / GIS tools
  - Robust for automated processing pipelines, including in polar regions and for antimeridian crossings
  - This is the # 1 top feature request from the NISAR Science Team for `nisarqa`
  - ISRO's SSAR RSLC products can now be processed through `nisarqa` to generate browse, reports, etc. (when file verification `.. is set to false.)

##  PR Review Strategy

  This is a large PR but it's structured logically:
  1. Infrastructure (utilities, parameters, grid methods) - foundational changes
  2. Product workflows (3 groups: backscatter, interferogram, offsets) - apply and use infrastructure consistently

I recommend reviewing in 3 passes:

  Pass 1: High-Level Understanding
  - Read PR description
  - Review screenshots and diagrams
  - Understand the two problems being solved
  - Check the before/after visual examples

  Pass 2: Core Infrastructure
  - Focus on foundational changes that all products use
  - These changes are reused across all 8 products

  Pass 3: Product-Specific Changes
  - Review how infrastructure is applied to each product family
  - Spot-check consistency across similar products

  See "Review Plan" section below for more detailed guidance.

  ## Changes Overview

  ### New Features
  - ✨ Optional EPSG 4326 browse PNG + KML generation for all products
  - ✨ Accurate KML corner coordinates computed from downsampled grids
  - ✨ Support for antimeridian crossing and polar regions
  - ✨ Proper handling of odd/even multilooking without half-pixel offsets

  ### Core Infrastructure Added
  - `BrowseOutputPaths` class for centralized management of browse file paths
  - `Browse4326ParamGroup` parameter classes (L1Radar/L2Geo variants) for centralized management of new runconfig parameters for generating the 4326 Browse.
  - `grid.downsample()` method with configurable downsampling modes
  - `grid.save_kml()` method for consistent KML generation from actual coordinate grids
  - `geocode_radar_raster()` - Geocodes range-Doppler grid rasters (float) to geocoded rasters; accommodates any EPSG.
  - `reproject_geo_raster()` - GDAL Warps geocoded rasters to a new EPSG
  - `_get_multilooked_center_coordinates()` utility for sub-pixel accuracy; handles odd/even nlooks correctly when multilooking

  ### Workflows Updated (All 8)
  **Level-1 (Radar)** - using ISCE3 geocoding:
  - RSLC, RIFG, RUNW, ROFF

  **Level-2 (Geocoded)** - using GDAL reprojection:
  - GSLC, GCOV, GUNW, GOFF

  ### File Naming Convention
  - Primary browse (no change): `BROWSE.png`, `BROWSE.kml`
  - EPSG 4326 browse: `BROWSE_4326.png`, `BROWSE_4326.kml`

  Output Files Generated:
  With output_browse_4326: false (default):
  - BROWSE.png - original coordinate system (accurate corners!)
  - BROWSE.kml - accurate corners

  With output_browse_4326: true:
  - BROWSE.png - original coordinate system (accurate corners!)
  - BROWSE.kml - accurate corners
  - BROWSE_4326.png - EPSG 4326 (lat/lon)
  - BROWSE_4326.kml - EPSG 4326 corners

  ### Backward Compatibility for Mission Operations
  *  **Fully backward compatible** - 4326 generation is opt-in (default: `false`)
  *  Original browse behavior unchanged when `output_browse_4326: false`. (Same output files generated)

  ## Technical Approach

  ### KML `latlonquad` Corner Accuracy Fix (Issue #2)
  ```python
  # OLD (Incorrect): Used full-resolution coordinates
  llq = compute_corners(full_resolution_coordinates)

  # NEW (Correct): Use downsampled coordinates matching browse image PNG. Example:
  decimated_coords = coords[::decimation_stride]
  llq = compute_corners(decimated_coords)
  ```

  ### Primary Browse + (optional) EPSG 4326 Generation pseudocode (Issue #1)

  For Level-1 (Radar) products:
  1. Multilook/decimate the source image to browse size
  2. Generate Primary PNG and compute accurate KML corners from downsampled radar grid
  3. Geocode browse-sized array to EPSG 4326 using ISCE3 with browse-sized RadarGrid
  4. Generate EPSG PNG and compute accurate KML corners from new geogrid

  For Level-2 (Geo) products:
  1. Multilook/decimate the source image to browse size
  3. Generate Primary PNG and compute accurate KML corners from downsampled geogrid
  3. Reproject browse-sized array to EPSG 4326 using GDAL
  4. Generate EPSG PNG and compute accurate KML corners from reprojected geogrid

  #### **Technical Diagram: Processing Flow**

```mermaid                                                                                                
  flowchart TD;
      A[Input NISAR Product] --> B[Load Source Raster<br/>Full Resolution];
      B --> C[Downsample<br/>to Primary Browse Size<br/>Track ky, kx];
      C --> D[Apply Image Correction<br/>gamma, clipping, etc.];
      D --> E[Create Downsampled Grid<br/>for Primary Browse];
      E --> F[Save Primary Browse.png<br/>Save Primary BROWSE.kml<br/>with Accurate Corners<br/> **Fix for Issue #2**];
      F --> G{optional:<br/>output_browse_4326 ?};
      G -->|yes| H[Geocode/Reproject to EPSG 4326<br/>- L1: ISCE3 geocoding<br/>- L2: GDAL
  reprojection<br/> **Fix for Issue #1** ];
      H --> I[Save BROWSE_4326.png<br/>Save BROWSE_4326.kml];
```

###  Functional Testing

  ✅ All 8 product types tested with real NISAR data
  ✅ Both frequencies (A and B) verified
  ✅ Multiple polarizations tested (SHNA, DHDH, QPDH, NASV)

###  Visual Validation

  ✅ KML overlays verified in Google Earth for accuracy
  ✅ Antimeridian crossing cases tested and working
  ✅ Polar regions with polar stereo projections validated
  ✅ Before/after comparison confirms corner accuracy improvements

###  Edge Cases

  ✅ Different EPSG codes (UTM zones, polar stereo)
  ✅ Antimeridian crossing
  ✅ Level 1 Browse Products with/without DEM files

  ## Suggested Sample Outputs & Screenshots

  ### Screenshots/Figures

**Available offline**

  #### F. **Example Runconfig Snippets**

  **F1: Minimal config (default behavior)**
  ```yaml
  runconfig:
    groups:
      qa_reports:
        # No browse_4326 section = default behavior (no 4326 output)
        backscatter_img:
          longest_side_max: 2048
```

  F2: Enable EPSG 4326 for L1 Radar (RSLC/RIFG/RUNW/ROFF)
```yaml
  runconfig:
    groups:
      qa_reports:
        browse_4326:               # yaml group is called `browse` for InSAR products
          output_browse_4326: true
          resample: biquintic          # ISCE3 interpolation method
```

  F3: Enable EPSG 4326 for L2 Geo (GSLC/GCOV/GUNW/GOFF)
```yaml
  runconfig:
    groups:
      qa_reports:
        browse_4326:              # yaml group is called `browse` for InSAR products
          output_browse_4326: true
          resample: average              # GDAL resampling method
```

##  Review Strategy for Teammates

 (copy-paste from above)
  This PR is large but logically structured. I recommend reviewing in 3 passes:

  Pass 1: High-Level Understanding
  - Read PR description
  - Review screenshots and diagrams
  - Understand the two problems being solved
  - Check the before/after visual examples
  - Review runconfig yamls, and output filenames

  Pass 2: Core Infrastructure
  - Focus on foundational changes that all products use
  - These changes are reused across all 8 products

  Pass 3: Product-Specific Changes
  - Review how infrastructure is applied to each product family
  - Spot-check consistency across similar products

###  Pass 1: High-Level Understanding

  Goal: Understand what problems are being solved and verify the approach

  Files to review:
  - PR description (this document)
  - Visual examples (screenshots A-F)
  - Technical diagram (processing flow)

  Questions to answer:
  - Do I understand Issue 1 (coordinate grid problem)?
  - Do I understand Issue 2 (corner accuracy problem)?
  - Does the visual evidence show both issues are fixed?
  - Is the approach (ISCE3 for L1, GDAL for L2) reasonable?

###  Pass 2: Core Infrastructure

  Goal: Review the foundational utilities that all products depend on

  Files:
  - Coordinate Grids: src/nisarqa/grids.py ⭐ NEW FILE
  - Geolocation functions: src/nisarqa/geolocation.py ⭐ NEW FILE
  - Centralized browse path management: src/nisarqa/parameters/browse_output_paths.py ⭐ NEW FILE
  - Runconfig updates: src/nisarqa/parameters/nisar_params.py (Browse4326ParamGroup classes)

#### Comments on design changes for: src/nisarqa/grids.py
Coordinate Grid, GeoGrid, and RadarGrid dataclasses:
       - These were moved from src/nisarqa/raster_classes.py to a new file, src/nisarqa/grids.py. 
       - Previously, these were inherited by GeoRaster and RadarRaster dataclasses. With this new PR, that inheritance design pattern becomes too cumbersome to maintain. The code has been updated to use a composition design pattern, where each Raster "has a" CoordinateGrid attribute.
       - These dataclasses were enhanced with several new methods, including downsample() and save_kml() methods.

###  Pass 3: Product-Specific Changes

  Goal: Verify infrastructure is applied correctly and consistently

  Review products in groups (they follow similar patterns):

####  Section 3A: Backscatter Products

  Products: RSLC, GSLC, GCOV
  Files:
  - src/nisarqa/workflows/rslc_qa.py
  - src/nisarqa/workflows/gslc_qa.py
  - src/nisarqa/workflows/gcov_qa.py
  - src/nisarqa/processing/non_insar/backscatter_images_and_browse.py

####  Section 3B: Interferogram Products

  Products: RIFG, RUNW, GUNW
  Files:
  - src/nisarqa/workflows/igram_qa.py
  - src/nisarqa/processing/insar/wrapped_phase_image_and_coh_mag.py
  - src/nisarqa/processing/insar/unwrapped_phase_image.py

####  Section 3C: Offset Products

  Products: ROFF, GOFF
  Files:
  - src/nisarqa/workflows/offsets_qa.py
  - src/nisarqa/processing/insar/az_and_slant_rng_offsets.py


src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+    Notes
+    -----
+    This is a private helper function. The file is NOT automatically cleaned up.
+    Use dem_file_manager() context manager for automatic cleanup.
+    """
+    # Create a uniquely-named file in the scratch directory
+    scratch = nisarqa.get_global_scratch_dir()
+    utc_now = datetime.now(timezone.utc).strftime("%Y%m%d%H%M%S%f")
+
+    dem_file = scratch / f"zero_height_dem_{utc_now}.tif"
```
bhawkins: I recommend removing the `utc_now` salt since there's no multithreading in QA so no chance of a race, and other processes each have their own scratch directory. If we want to make it thread-safe for future compatibility, with should use the OS mkstemp.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+
+    # 1. Define resolution
+    res = 1.0  # 1 degree resolution
+
+    # Calculate dimensions (360x180 for 1-degree resolution)
+    width = int(360 / res)
+    height = int(180 / res)
```
bhawkins, line 104: Do we need +1 on these to fully cover the map? I guess gaps don't matter too much since the DEMInterpolator will ultimately just use a reference value of zero to fill points that fall outside the bounds.

Suggest running `gdalinfo` on the output TIFF file to verify coverage bounds.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+    to geographic/projected coordinates.
+
+    Parameters
+    ----------
+    radar_array : array-like
+        Input range-Doppler image array (e.g. NISAR Level-1) to be geocoded.
+        Must be real-valued (float).
+    radargrid : isce3.product.RadarGridParameters
+        ISCE3 radargrid parameters specifying the range-Doppler grid associated
+        with the input `radar_array`.
```
bhawkins:
```suggestion
    radargrid : nisarqa.RadarGrid
        Radar grid parameters specifying the range-Doppler grid associated
        with the input `radar_array`.
```

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+        )
+
+    # Use scratch directory with temporary files for ISCE3 raster I/O.
+    # GDAL's in-memory option has potential security issues.
+    scratch = nisarqa.get_global_scratch_dir()
+
+    gdal_dtype = gdal.GDT_Float64
+    output_dtype = np.float64
+    # Create a uniquely-named string for filenaming
+    utc_now = datetime.now(timezone.utc).strftime("%Y%m%d%H%M%S%f")
```
bhawkins: `mkstemp` or nothing

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+    Recommend only using it to geocode relatively small rasters, such as
+    browse image arrays. To geocode full-size NISAR rasters, suggest using
+    ISCE3 directly.
+    """
+
+    if np.iscomplexobj(radar_array):
+        raise ValueError(
+            f"{type(radar_array)=} which is complex-valued. Only real-valued"
+            " data currently supported."
+        )
```
bhawkins, line 199: The rest of this function uses real-valued data types, but nowadays this isn't required by geocode_cov in isce3. So it could be made more generic.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+                gdal_dtype,
+            )
+
+            # Ensure float type
+            raster_array = radar_array.astype(np.float64)
+
+            # Get lon/lat corners from radar grid
+            llq = radargrid.get_latlonquad(
+                orbit=orbit,
+                wavelength=wavelength,
```
bhawkins: Why is there `wavelength` here? Only need it if we have a Doppler.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+
+            width = maxx - minx
+            height = maxy - miny
+
+            # Determine resolution based on the longest side of the array.
+            # Handle width and height independently. For example, in lon/lat
+            # coordinates, 1 degree of latitude is not equivalent to 1 degree
+            # of longitude, particularly towards the polar regions.
+            resolution_x = width / max(np.shape(radar_array))
+            resolution_y = height / max(np.shape(radar_array))
```
bhawkins: I think we should revisit this formula. In our discussion, it turns out we're working towards two goals:
* The output image size needs to be <= 2048 pixels along each dimension.
* The output image should have similar spatial resolution as the input image.

The input image satisfies the 2048 pixel constraint, so dividing the x-span and y-span by the max dimension size also enforces it on the output. However, it doesn't preserve the aspect ratio of the spacing in x and y. So we propose something like the following:

```Python
# Want output dimensions <= max input dimension.
maxdim = max(radar_array.shape)
a = 6378137.
if epsg == 4326:  # latlon dx/dy varies by cos(lat)
    # compute spacing in meters
    dx_m = width / maxdim * a / cos(lat)
    dy_m = height / maxdim * a
    spacing_m = max(dx_m, dy_m)
    resolution_x = spacing_m / (a / cos(lat))
    resolution_y = spacing_m / a
else:
    # Assume polar stereo or UTM, which have spacing in m
    # so dx/dy = 1.  Should put this assumption in docstring.
    dx_m = width / maxdim
    dy_m = height / maxdim
    resolution_x = resolution_y = max(dx_m, dy_m)
```

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+                x, y, z = proj.forward([lon_rad, lat_rad, 0])
+                corners_proj.append((x, y))
+
+            # Get bounding box in target projection's native units
+            xs = [c[0] for c in corners_proj]
+            ys = [c[1] for c in corners_proj]
+            minx, maxx = min(xs), max(xs)
+            miny, maxy = min(ys), max(ys)
+
+            width = maxx - minx
```
bhawkins: This can wrap at the dateline.
```suggestion
            width = maxx - minx
            if epsg == 4326 and width > np.pi:
                xs = [xi for xi in xs if xi > 0 else xi + (2 * np.pi)]
                width = max(xs) - min(xs)
```
Suggest testing this change with the dateline test case you already have near the pole as well as a case like Fiji closer to the equator.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+                output_file,
+                geocode_obj.geogrid_width,
+                geocode_obj.geogrid_length,
+                1,
+                gdal_dtype,
+            )
+
+            # Set output file's projection and geotransform
+            srs = osr.SpatialReference()
+            srs.ImportFromEPSG(epsg)
```
bhawkins: Note that since we're using a `SpatialReference` object we can use the `srs.IsGeographic()` method to condition the lat/lon cases vs the other projections instead of on the epsg directly.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,581 @@
...
+                radar_grid=isce3_radargrid,
+                input_raster=input_raster,
+                output_raster=output_raster,
+                dem_raster=dem,
+                output_mode=isce3.geocode.GeocodeOutputMode.INTERP,
+            )
+
+            # Explicitly close the ISCE3 Rasters
+            input_raster = None
+            output_raster = None
```
bhawkins: More GDAL quirks we have to work around.
```suggestion
            output_raster.close_dataset()
            output_raster = None
```

bhawkins (commented):
Comments from today's online meeting.

src/nisarqa/processing/geolocation.py
bhawkins: After writing this comment, we looked at the icse3 code and it turns out that currently all the necessary type overloads have python bindings, but they all have different names. So we'd probably want to add a generic wrapper to isce3 that handles the type dispatch under the hood. (I'd recommend that this wrapper also implement a proper constructor, but that's another story.) Once that's released then nisarqa could get rid of the hard-coded Float64 in this method, which would expand our options for reprojecting wrapped phase layers.

src/nisarqa/processing/geolocation.py
nemo794: Hmm, good question. As suggested, I ran `gdalinfo`, and the code as-is provides full coverage: [-180, 180] and [-90,90]. So, I won't make any changes at this time.

But in case there's something subtle that I didn't catch (e.g. the center is not exactly at 0, 0), here's the `gdalinfo` output to double-check:

```
(qa310) niemoell@MT-316822 nisarqa % gdalinfo /Users/niemoell/Desktop/qa_public/nisarqa/scratch/qa-rifg-20260508213937-lp1vf8x9/zero_height_dem_20260508213941082213.tif 
Driver: GTiff/GeoTIFF
Files: /Users/niemoell/Desktop/qa_public/nisarqa/scratch/qa-rifg-20260508213937-lp1vf8x9/zero_height_dem_20260508213941082213.tif
Size is 360, 180
Coordinate System is:
GEOGCRS["WGS 84",
    ENSEMBLE["World Geodetic System 1984 ensemble",
        MEMBER["World Geodetic System 1984 (Transit)"],
        MEMBER["World Geodetic System 1984 (G730)"],
        MEMBER["World Geodetic System 1984 (G873)"],
        MEMBER["World Geodetic System 1984 (G1150)"],
        MEMBER["World Geodetic System 1984 (G1674)"],
        MEMBER["World Geodetic System 1984 (G1762)"],
        MEMBER["World Geodetic System 1984 (G2139)"],
        MEMBER["World Geodetic System 1984 (G2296)"],
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]],
        ENSEMBLEACCURACY[2.0]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
        SCOPE["Horizontal component of 3D system."],
        AREA["World."],
        BBOX[-90,-180,90,180]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
Origin = (-180.000000000000000,90.000000000000000)
Pixel Size = (1.000000000000000,-1.000000000000000)
Metadata:
  AREA_OR_POINT=Area
Image Structure Metadata:
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  (-180.0000000,  90.0000000) (180d 0' 0.00"W, 90d 0' 0.00"N)
Lower Left  (-180.0000000, -90.0000000) (180d 0' 0.00"W, 90d 0' 0.00"S)
Upper Right ( 180.0000000,  90.0000000) (180d 0' 0.00"E, 90d 0' 0.00"N)
Lower Right ( 180.0000000, -90.0000000) (180d 0' 0.00"E, 90d 0' 0.00"S)
Center      (   0.0000000,   0.0000000) (  0d 0' 0.01"E,  0d 0' 0.01"N)
Band 1 Block=360x11 Type=Int16, ColorInterp=Gray
```

src/nisarqa/processing/geolocation.py
nemo794: Updated to `mkstemp`. Simple enough to correctly now, vs. dealing with a race condition bug later.

src/nisarqa/processing/geolocation.py
nemo794: Good call -- decided to put these in a context manager with NameTemporaryFile(..), to allow for unique filenames.

TBH, this makes the code a bit more elegant:
* We can have these temp files follow the same context-manager design pattern used for the DEM file -- a triple context manager!
* We can now remove the try-catch handling
* We decrease the indent by one level on that large code block!

🚀

src/nisarqa/processing/geolocation.py
nemo794: Good discussion -- documented in https://github.com/isce-framework/nisarqa/issues/162

src/nisarqa/processing/geolocation.py
nemo794: As discussed, the original PR's PNGs did not exhibit any issues when crossing the dateline. I figured out why, and updated the comments to explain why in this commit: https://github.com/isce-framework/nisarqa/pull/161/commits/fe286e42366c624847d60126c1eb7f3818be05ef

bhawkins: Thanks for tracking that down. It makes sense.

This got me thinking about situations where the unwrapping trick fails. On a sphere of radius $a$, a longitude step $\Delta \lambda$ subtends an arc length $L = \Delta\lambda a \cos\phi$ where $\phi$ is the latitude. The basic constraint for unwrapping is that the longitude span over any dimension is less than 180°. So if we pick a length scale $L = \sqrt{2} \cdot 240 \rm{km}$ then we can solve for the latitude $\phi = \cos^{-1} \frac{L}{\pi a}$ which works out to around ±89°. So we're safe as long as we never get within about one degree of the poles or process bigger frames.

src/nisarqa/processing/geolocation.py
nemo794: Very helpful discussion and algorithm - thank you! It seems that `proj.forward(..)` returns lat/lon in units of degrees (not radians), so I adjusted the proposed algorithm accordingly. But, the basic idea of the algorithm is the same. 🚀

nemo794: Also, when testing again, the margin that ISCE3's `update_geogrid()` applies seems to be dynamically computed, which led to the latlon browse images having dimensions that are larger than the max dim of the input L1 raster. :(

So, I added a quick binary search to find the optimal margin; this allows the algorithm to respect all of the goals stated above.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,658 @@
...
+            lat_rad = np.deg2rad(corner.lat)
+            # Forward transform: lon/lat -> target projection
+            # Returns coordinates in projection's native units
+            # (degrees for EPSG 4326 lat/lon, meters for UTM, etc.)
+            # Note: for EPSG 4326, will not re-wrap longitudes to [-180, 180].
+            x, y, z = proj.forward([lon_rad, lat_rad, 0])
+            corners_proj.append((x, y))
+
+        # Get bounding box in target projection's native units
+        xs = [c[0] for c in corners_proj]
```
bhawkins, line 276:
```suggestion
        xs = [c[0] for c in corners_proj]
        if srs.IsGeographic:
            xs = unwrap_longitudes(xs)
```
While the `get_latlonquad` method unwraps the longitudes, I'm not sure we should trust `proj.forward` not to wrap them again. This way the code would be robust regardless of `proj` behavior.

Edit: When I wrote this, I'd assumed that `proj` was from the [proj library](https://proj.org/en/stable/) which we don't have much control over. However, on my second look I see it's from isce3 and it's easy to verify [here](https://github.com/isce-framework/isce3/blob/df9af9a7ebfcef5ab39a6c66f01310ba3d5b2bde/cxx/isce3/core/Projections.h#L127) that there's no interval normalization. So I'm less concerned now.

It also means we're only dealing with the small subset of EPSG codes supported by isce3. The docstring says "any valid EPSG" which is only true if "valid" is understood to mean "supported by `isce3::core::createProj`" and might be worth clarifying.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,658 @@
...
+
+        # Create a SpatialReference object to check output projection type
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(epsg)
+
+        # ISCE3's update_geogrid() dynamically adds a margin when determining
+        # the geogrid to ensure that the swath's corners are all contained.
+        # To compensate and meet the maxdim constraint, use binary search to
+        # find the smallest margin that produces output dimensions < maxdim.
+        # As margin increases, target_dim decreases, resolution increases,
```
bhawkins:
```suggestion
        # As margin increases, target_dim decreases, resolution gets coarser,
```

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,658 @@
...
+            )
+            geo_obj.update_geogrid(isce3_radargrid, dem)
+
+            return geo_obj
+
+        # Binary search for optimal margin (range: 0 to 200 pixels)
+        left, right = 0, 200
+        geocode_obj = None
+
+        while left <= right:
```
bhawkins: I think we can do this with a single iteration step. Basically the algorithm would be

1. Compute dx and dy using above algorithm.
2. Call `geogrid(dx, dy, ...)  # other parameters defaulted`
3. Call `update_geogrid`, which may enlarge the image due to the margin. However, now the corner coordinates are known.
4. Recompute dx and dy using known corner coordinates and `maxdim`.
5. Call `geogrid(dx, dy, x_start, y_start, width, length)`.

This avoids unnecessary iteration and having to guess about `left` and `right` bounds that'll work in all cases.

src/nisarqa/processing/geolocation.py
```diff
@@ -0,0 +1,658 @@
...
+            "dstNodata": fill_value,
+            # Constrain output dimensions to match input dimensions.
+            # This is intentional for browse image generation to ensure
+            # predictable output shapes and improve visual appearance
+            # for the browse products (particularly in polar regions),
+            # regardless of the source and target projections.
+            # This means that the output's x spacing and y spacing will no
+            # longer be approx. equal (i.e. output won't have ~square pixels).
+            "height": src_height,
+            "width": src_width,
```
bhawkins: I'll try to come up with a suggestion to improve this.

bhawkins: Suggestion here https://github.com/nemo794/nisarqa/pull/6

src/nisarqa/processing/geolocation.py
nemo794: Comment no longer relevant -- it is superseded by https://github.com/isce-framework/nisarqa/pull/161#discussion_r3236030769

src/nisarqa/processing/geolocation.py
nemo794: Addressed in https://github.com/isce-framework/nisarqa/pull/161/commits/2dea4cf90045854e5cefe0936a2ecb55cb8fd95a . Unexpectedly, the ISCE3 io Raster for the DEM file kept getting corrupted mid-run during testing (it caused segmentation faults, etc.), so this new algorithm gets around that bug by opening and closing the ISCE io Raster for the DEM for each use. (Note: This wasn't an issue previously.)

src/nisarqa/grids.py
```diff
@@ -0,0 +1,896 @@
...
+        Downsample grid by the given strides and mode.
+
+        Parameters
+        ----------
+        y_stride, x_stride : int
+            Stride for downsampling along the Y (azimuth) or X (slant range)
+            axis, respectively.
+        mode : str, optional
+            Downsampling algorithm. One of:
+                "decimate" : (default) Pure decimation. For example, if
```
bhawkins:
```suggestion
                "decimate" : Pure decimation. For example, if
```

src/nisarqa/grids.py
```diff
@@ -0,0 +1,896 @@
...
+            y_post, y_coords[1] - y_coords[0]
+        ), f"{y_post=}, {y_coords[1] - y_coords[0]=}"
+
+        x_post = self.x_posting * x_stride
+        assert np.isclose(x_post, x_coords[1] - x_coords[0])
+
+        return y_coords, x_coords, y_post, x_post
+
+
+@dataclass
```
bhawkins, line 207: Can this be made frozen? Then there's no way that attributes can get out of sync (e.g., coordinate vectors vs the start/stop edges). This could be done in a separate PR.

src/nisarqa/grids.py
```diff
@@ -0,0 +1,896 @@
...
+
+        # Convert look_side to ISCE3 LookSide enum if it's a string
+        # TODO - check if it is ok to skip conversion and pass the string
+        if isinstance(look_side, str):
+            if look_side.lower() == "left":
+                look_side = isce3.core.LookSide.Left
+            elif look_side.lower() == "right":
+                look_side = isce3.core.LookSide.Right
+            else:
+                raise ValueError(f"Invalid look_side string: {look_side}")
```
bhawkins: isce3 has a constructor that takes a `look_side::str` argument, so this shouldn't be necessary.

src/nisarqa/grids.py
```diff
@@ -0,0 +1,896 @@
...
+        corner_lons = []
+        for corner in corners:
+            lon_rad, lat_rad, h = proj.inverse(corner)
+            lon_deg = np.rad2deg(lon_rad)
+            corner_lons.append(lon_deg)
+
+        # Check if any longitude extends beyond [-180, 180]
+        for lon in corner_lons:
+            if lon < -180 or lon > 180:
+                return True
```
bhawkins: Is this really necessary? They could all be in the range [183, 186], in which case they're all > 180 and would cause this to return `True` even though there's no antimeridian crossing.

src/nisarqa/grids.py
```diff
@@ -0,0 +1,896 @@
...
+
+        # Check for discontinuous jumps between corner longitudes
+        # A jump > 180 indicates wrapping across the dateline
+        for i in range(len(corner_lons)):
+            for j in range(i + 1, len(corner_lons)):
+                if abs(corner_lons[i] - corner_lons[j]) > 180:
+                    return True
```
bhawkins: Could probably do this with `max(lons) - min(lons)` but there are only 6 checks here, so it's probably about the same.

src/nisarqa/raster_classes.py
```diff
@@ -948,6 +609,17 @@ def compare_raster_metadata(raster1, raster2, almost_identical=True):
...
+            for grid_f in fields(r1_val):
+                field_name = grid_f.name
+                grid1_val = getattr(r1_val, field_name)
+                grid2_val = getattr(r2_val, field_name)
+            if grid1_val != grid2_val:
+                raise ValueError(
+                    f"Values do not match for Coordinate Grid's `{field_name}`"
+                    f" field. `raster1`'s grid has value {r1_val},"
+                    f" but `raster2`'s grid has value {r2_val}."
+                )
```
bhawkins: Need to indent this block. As is, we're only checking the last field.

src/nisarqa/input_product_readers/gcov_reader.py
```diff
@@ -37,6 +37,36 @@ def get_pols(self, freq: str) -> tuple[str, ...]:
 
         return pols
 
+    def center_freq(self, freq: str) -> float:
```
bhawkins: It seems like GCOV and GSLC both have this field in the source data group, so this method could probably go up a level in the class hierarchy.

bhawkins (commented):
Lots of progress today!

src/nisarqa/grids.py
nemo794: Good question -- in practice, `nisarqa` treats the CoordinateGrid dataclasses as frozen (meaning, after instantiation, `nisarqa` does not modify instances during runtime.)

Making this change is a bit out-of-scope for this PR because it would require updating the `__post_init__` methods. So, I've documented it as an Issue: https://github.com/isce-framework/nisarqa/issues/163

src/nisarqa/grids.py
nemo794: Fun fact: It appears that if the look side is a `str`, then the parameter name is `look_side`. If the look side is an `isce3.core.LookSide`, then the parameter name is `lookside` (no underscore).

I'll update QA to explain this, but generally leave the logic as-is.

Commit where this update is implemented: https://github.com/isce-framework/nisarqa/pull/161/commits/a384fd1e936a17d8c0e646628137d521464f9cb0

Here's the traceback:

```
    radar_grid = isce3.product.RadarGridParameters(
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. isce3.ext.isce3.product.RadarGridParameters(h5file: str, freq: str = 'A')
    2. isce3.ext.isce3.product.RadarGridParameters(sensing_start: typing.SupportsFloat, wavelength: typing.SupportsFloat, prf: typing.SupportsFloat, starting_range: typing.SupportsFloat, range_pixel_spacing: typing.SupportsFloat, lookside: isce3.ext.isce3.core.LookSide, length: typing.SupportsInt, width: typing.SupportsInt, ref_epoch: isce3.ext.isce3.core.DateTime)
    3. isce3.ext.isce3.product.RadarGridParameters(sensing_start: typing.SupportsFloat, wavelength: typing.SupportsFloat, prf: typing.SupportsFloat, starting_range: typing.SupportsFloat, range_pixel_spacing: typing.SupportsFloat, look_side: str, length: typing.SupportsInt, width: typing.SupportsInt, ref_epoch: isce3.ext.isce3.core.DateTime)

Invoked with: kwargs: sensing_start=22200.446052631578, wavelength=0.2360571, prf=138.1818181818182, starting_range=743467.339520901, range_pixel_spacing=18.737028625, lookside='right', length=1959, width=1543, ref_epoch=2008-11-27T00:00:00.000000000
```

nemo794: FYI -- the look side value gets passed around a lot in `nisarqa`, but the only other places in `nisarqa` where the look side is used for actual useful work are calls to ISCE3's `isce3.geometry.rdr2geo_bracket(..., side=.., ...)`.

Digging into ISCE3, it looks like `r2r2geo_bracket(..)` handles the normalization of the various input types via a call to `duck_look_side()` (call is here: https://github.com/isce-framework/isce3/blob/df9af9a7ebfcef5ab39a6c66f01310ba3d5b2bde/python/extensions/pybind_isce3/geometry/rdr2geo_roots.cpp#L27 ).

Which means those calls in `nisarqa` should (happily) be ok.

src/nisarqa/processing/geolocation.py
nemo794: Suggestion reviewed and merged into this branch, here: https://github.com/isce-framework/nisarqa/pull/161/commits/b58ae3d95c1396aa8b3c02222404cfa7f7aefab0

Thanks, @bhawkins !

src/nisarqa/grids.py
nemo794: Good catch! The first check was could lead to false positives -- removed!

I initially removed just the first check, but then realized we could also have the opposite problem (false negatives) similar to this Issue in ISCE3: https://github.com/isce-framework/isce3/issues/117 .

For example, a grid with corners at [179, 181, 179, 181] (where 181° is unwrapped, representing -179°) would have no pairwise differences > 180° and would incorrectly return False.

So, I updated the function to wrap all longitudes to [-180, 180) before checking for discontinuities.

Commit with fix: https://github.com/isce-framework/nisarqa/pull/161/commits/e0c0cd780fd7795a90db0604f4d5187d6d61f93f

src/nisarqa/raster_classes.py
nemo794: Commit with fix: https://github.com/isce-framework/nisarqa/pull/161/commits/a4fb46a3fb3accc9c77994196f90bbd03b637fd7

src/nisarqa/input_product_readers/gcov_reader.py
nemo794: Commit with fix: https://github.com/isce-framework/nisarqa/pull/161/commits/bf0ed8d8b2ff99421076dfe0298e2480339dece9

src/nisarqa/grids.py
bhawkins: Let's open an issue on the isce3 repo about the spelling discrepancy.

src/nisarqa/input_product_readers/gslc_reader.py
```diff
@@ -2,6 +2,8 @@
 
 from dataclasses import dataclass
 
+import h5py
+
```
bhawkins: Can probably remove this

src/nisarqa/processing/insar/_utils.py
```diff
@@ -0,0 +1,422 @@
...
+    Returns
+    -------
+    phs_img : numpy.ndarray
+        Phase image (float-valued). For complex input, this is the result
+        of np.angle(). For real input, this is a copy of the data.
+    """
+    phs_img = np.array(phs_or_complex_raster.data, copy=True)
+
+    if phs_or_complex_raster.is_complex:
+        phs_img = np.angle(phs_img)
```
bhawkins:
```suggestion
    if phs_or_complex_raster.is_complex:
        phs_img = np.angle(phs_img)
    else:
        phs_img = np.array(phs_or_complex_raster.data, copy=True)
```

src/nisarqa/parameters/browse_output_paths.py
```diff
@@ -0,0 +1,203 @@
...
+        >>> paths.get_browse_filename()
+        'BROWSE.png'
+        >>> paths.get_browse_filename(suffix="A_HH")
+        'BROWSE_A_HH.png'
+        >>> paths.get_browse_filename(suffix="LATLON")
+        'BROWSE_LATLON.png'
+        """
+        if suffix is None:
+            return self.browse_filename
+        return f"{self.browse_stem}_{suffix}.png"
```
bhawkins:
```suggestion
        return f"{self.browse_stem}{suffix}.png"
```
Maybe put the underscore in the global LATLON_SUFFIX instead so a motivated user could hack the suffix to the empty string and have the LATLON variants clobber the other ones.

src/nisarqa/parameters/browse_output_paths.py
```diff
@@ -0,0 +1,203 @@
...
+        >>> paths.get_kml_filename()
+        'BROWSE.kml'
+        >>> paths.get_kml_filename(suffix="A_HH")
+        'BROWSE_A_HH.kml'
+        >>> paths.get_kml_filename(suffix="LATLON")
+        'BROWSE_LATLON.kml'
+        """
+        if suffix is None:
+            return self.kml_filename
+        return f"{self.kml_stem}_{suffix}.kml"
```
bhawkins: Ditto clobber option.

bhawkins (commented):
One concern is that there should be an option to not generate the KML corresponding to the PNG in the original coordinate system.

src/nisarqa/grids.py
nemo794: Issue raised: https://github.com/isce-framework/isce3/issues/272

src/nisarqa/processing/insar/_utils.py
nemo794: Updated in https://github.com/isce-framework/nisarqa/pull/161/commits/dd5d0caaf64dd62211867b8648373b11adedbe96

src/nisarqa/input_product_readers/gslc_reader.py
nemo794: Removed unused imports: https://github.com/isce-framework/nisarqa/pull/161/commits/06d921e912a34b3fe40d579f2f4f577b6a6ab3b9

src/nisarqa/parameters/browse_output_paths.py
nemo794: Now hackable: https://github.com/isce-framework/nisarqa/pull/161/commits/91915b2a1715395503e6b58c566c514b031990c7

src/nisarqa/parameters/nisar_params.py
```diff
@@ -988,6 +988,177 @@ def at_least_one_wkflw_requested(self):
...
+            "yaml_attrs": YamlAttrs(
+                name="output_browse_latlon",
+                descr=f"""True to generate a version of the primary browse PNG+KML that is
+        projected to EPSG 4326 (lon/lat) coordinate system for accurate
+        geolocation in GIS software.
+        False to only generate primary browse outputs.
+        The EPSG 4326 PNG+KML version will be in addition to the primary
+        browse PNG+KML outputs (where the PNG reflects the input product's
+        native coordinate system). The EPSG 4326 outputs will use the
+        suffix '_{nisarqa.LATLON_SUFFIX}' (e.g., 'BROWSE_{nisarqa.LATLON_SUFFIX}.png' and 'BROWSE_{nisarqa.LATLON_SUFFIX}.kml').""",
```
bhawkins:
```suggestion
        suffix '{nisarqa.LATLON_SUFFIX}' (e.g., 'BROWSE{nisarqa.LATLON_SUFFIX}.png' and 'BROWSE{nisarqa.LATLON_SUFFIX}.kml').""",
```

src/nisarqa/processing/insar/quiver_plots.py
```diff
@@ -923,6 +1042,35 @@ def get_offset_values_in_projected_coordinates(
             # object we created above.
             x_shift, y_shift, _ = proj.forward([lon, lat, 0])
```
bhawkins:
```suggestion
            x_shifted, y_shifted, _ = proj.forward([lon, lat, 0])
```
When I read `x_shift` I thought it was the amount of shift (e.g., Δx) rather than the shifted coordinate. That made it hard to understand the logic below.

bhawkins (approved):
LGTM

src/nisarqa/input_product_readers/non_insar_geo_product.py
```diff
@@ -13,33 +14,93 @@
...
 
-        with self.get_raster(freq=freq, pol=pol) as img:
-            x_start = img.x_start
-            x_stop = img.x_stop
+        Returns
+        -------
+        source_data_path : str
+            Path to the sourceData group.
+                Standard Format:
+                    "/science/<band>/<product_type>/metadata/sourceData"
```
gshiroma:
```suggestion
                    "/science/<instrument_name>/<product_type>/metadata/sourceData"
```

src/nisarqa/input_product_readers/non_insar_geo_product.py
```diff
@@ -13,33 +14,93 @@
...
+    def _source_data_swaths_path(self) -> str:
+        """
+        Get the path to the sourceData/swaths group.
+
+        Returns
+        -------
+        source_data_path : str
+            Path to the sourceData group.
+                Standard Format:
+                    "/science/<band>/<product_type>/metadata/sourceData/swaths"
```
gshiroma:
```suggestion
                    "/science/<instrument_name>/<product_type>/metadata/sourceData/swaths"
```

src/nisarqa/input_product_readers/non_insar_geo_product.py
```diff
@@ -13,33 +14,93 @@
...
+        ----------
+        freq : str
+            Must be either "A" or "B".
+
+        Returns
+        -------
+        source_data_path : str
+            Path to the sourceData group.
+                Standard Format:
+                    "/science/<band>/<product_type>/metadata/sourceData/swaths/frequency<freq>"
```
gshiroma:
```suggestion
                    "/science/<instrument_name>/<product_type>/metadata/sourceData/swaths/frequency<freq>"
```

gshiroma (approved):
We reviewed this PR in several meetings. It was decided that there will be a separate PR to enable or disable the generation of KML files independently from the PNG files.

src/nisarqa/input_product_readers/non_insar_geo_product.py
nemo794: Good catch, @gshiroma . This distinction between "band" vs. "instrument name" (e.g. "L" vs. "LSAR") is subtle but important. I've gone through the codebase to ensure `nisarqa` is internally consistent in its usage: https://github.com/isce-framework/nisarqa/pull/161/commits/691b59ae1414e30036df8c86b08bbd4ef9cb23e9

fwiw, for brevity I used "instrument" instead of "instrument_name"... the lines were getting very long, and in the doctsrings always include an example (e.g. "/science/LSAR/RSLC/..."), so the meaning should be clear to users.

src/nisarqa/input_product_readers/non_insar_geo_product.py
nemo794: See: https://github.com/isce-framework/nisarqa/pull/161#discussion_r3313344004

src/nisarqa/input_product_readers/non_insar_geo_product.py
nemo794: See: https://github.com/isce-framework/nisarqa/pull/161#discussion_r3313344004

src/nisarqa/processing/insar/quiver_plots.py
nemo794: Updated in: https://github.com/isce-framework/nisarqa/pull/161/commits/9ecc1e69304ad1c6bab6990c2384e6dc352d9174

Tyler-g-hudson (approved):
LGTM based on the review @bhawkins very helpfully led offline. Any comments of mine have already been addressed above.

PR #164, opened 2026-05-27
Add `_NATIVE` suffix to default browse products

nemo794:
While implementing #161 , there arose much discussion about the naming for the browse PNG+KML pairs. That PR simply added a "_LATLON" suffix to the new files, but keeping the original PNG+KML pairs unlabeled seemed to lack easy grok-ability for casual users. Instead, we want to implement a filenaming convention with a focus on clarity, so that the users could know immediately what they were viewing/opening.

 After discussion with  Marco @hfattahi @bhawkins @gshiroma , the suggested path forward from that meeting:

* For PNGs that represent the science HDF5's native coordinate system (range-Doppler, UTM, or polar-stereo), add the suffix "_NATIVE"
* For PNGs whose contents have been projected to EPSG 4326, use the suffix "_LATLON"
* Provide only the KML for the "_LATLON.png", and keep the suffix "_LATLON" for the kml. This makes it extremely clear to users that the KML only pairs with the reprojected PNG.

Given that, here's what an RSLC SHNA granule's outputs would ideally be, as presented to the user in ASF Vertex:

```
<granule ID>_NATIVE_A_HH.png
<granule ID>_NATIVE.png
<granule ID>_LATLON.png
<granule ID>_LATLON.kml
<granule ID>_QA_STATS.h5
<granule ID>_QA_REPORT.pdf
<granule ID>_QA_LOG.txt
<granule ID>_QA_SUMMARY.csv
```

However, ADT and advanced users will likely want to still have access to the KMLs that pair with the NATIVE PNGs. This is particularly useful for Level-1 products on the range-Doppler grid where issues often arise along either the slant range axis or the along-track axis (independently), and geocoding to lat/lon might mask these issues.

So, this PR simply updates the naming convention to include "native". As a bonus, this should help users grok why the native browse images are not 100% accurately geolocated in Google Earth.

With this PR, the outputs direct from `nisarqa` for an RSLC SHNA granule with individual PNGs turned on and LATLON turned on would be:

```
BROWSE_NATIVE_A_HH.png
BROWSE_NATIVE_A_HH.kml
BROWSE_NATIVE.png
BROWSE_NATIVE.kml
BROWSE_LATLON.png
BROWSE_LATLON.kml
STATS.h5
BROWSEREPORT.pdf
LOG.txt
SUMMARY.csv
```
(outputs will be similar for other products)


bhawkins (approved):
LGTM

Tyler-g-hudson (approved):
LGTM!

PR #165, opened 2026-05-28
Add Product Spec XMLs v1.5.0

nemo794:
There was a new PIX Release v1.5.0.

This PR adds those new specs to `nisarqa`, for use by the XML Checker.

Reviewers: Please double-check that the latest-and-greatest PIX XMLs were correctly copied over. The last PIX PR to be merged was https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/357 ; if those changes are reflected in these XMLs, then we should be good.

Thanks!


xhuang-jpl (approved):
I have checked the InSAR XMLs, they are consistent with the PIX XMLs. Thank you @nemo794

Tyler-g-hudson (approved):
LGTM; I have confirmed using a diff checker that these files and the ones in PIX are identical.
Easy, casual 19,000 line PR!

